PR #102, opened 2019-03-09
Log traffic secrets for use with Wireshark

jlaine:
picotls's `setup_traffic_protection` logs `CLIENT_*` and `SERVER_*` if secret
logging is enabled, except if a traffic key callback is specified.

We can leverage this to log the secrets on the quicly side with a QUIC_
prefix, allowing full decoding with Wireshark's HEAD version.


jlaine:
This works for me:

image

Feel free to let me know if the code matches the expected style.

kazuho:
Thank you for the PR.

I've update picotls (that unfortunately introduced an API incompatibility) as well as updating the code to cover that.

PR #306, opened 2020-03-25
Add initial implementation of NEW_CONNECTION_ID handling

hfujita:
This PR implements an initial implementation of NEW_CONNECTION_ID frame handling, along with infrastructure to send NEW_CONNECTION_ID and RETIRE_CONNECTION_ID frames.

WIth this PR, quicly issues extra CIDs until it fills peer's active_connection_id_limit upon connection establishment. However these extra CIDs will never be used as of this patch.


hfujita:
I think this PR is ready for review. As of now no new test cases were added. I manually verified that NEW_CONNECTION_ID frames were actually sent and received.

lib/quicly.c
```diff
@@ -3431,6 +3627,23 @@ static int update_traffic_key_cb(ptls_update_traffic_key_t *self, ptls_t *tls, i
...
+    if (path_id_limit > QUICLY_MAX_PATH_ID)
+        path_id_limit = QUICLY_MAX_PATH_ID;
+
+    while (conn->egress.new_cid.max_path_id < path_id_limit) {
+        int ret = schedule_new_connection_id(conn, conn->egress.new_cid.max_path_id);
+        if (ret != 0)
+            return ret;
+        conn->egress.new_cid.max_path_id++;
+    }
+
```
kazuho: I think we should be scheduling the emission of NCID frames when:
* we obtain 1-RTT write key (see the `if` right above)
* when we receive RCID

include/quicly.h
```diff
@@ -431,6 +435,24 @@ struct st_quicly_default_scheduler_state_t {
...
+ * records an extra CID given by the peer via NEW_CONNECTION_ID frame
+ */
+struct st_quicly_spare_cid_t {
+    /**
+     * indicates whether this record holds an active (given by peer and not retired) CID
+     */
+    int is_active;
+    uint64_t sequence;
+    quicly_cid_t cid;
+    struct st_quicly_stateless_reset_t stateless_reset;
```
kazuho: Let's use `uint8_t [QUICLY_STATELESS_RESET_TOKEN_LEN]` here. Please refer to my comment below for the rationale.

include/quicly.h
```diff
@@ -470,16 +493,25 @@ struct _st_quicly_conn_public_t {
         /**
          * stateless reset token corresponding to the CID
          */
-        struct {
-            uint8_t *token;
-            uint8_t _buf[QUICLY_STATELESS_RESET_TOKEN_LEN];
-        } stateless_reset;
+        struct st_quicly_stateless_reset_t stateless_reset;
```
kazuho: I do not think we need this generalization. The reason we have `*token` and `_buf` is because we might not have stateless reset tokens for the first CID. That's unique to the first CID, and therefore we do not need to generalize it.

include/quicly.h
```diff
@@ -457,6 +479,7 @@ struct _st_quicly_conn_public_t {
          */
         quicly_cid_t offered_cid;
         struct st_quicly_conn_streamgroup_state_t bidi, uni;
+        uint32_t n_retired_cids;
```
kazuho: I believe it is in your ToDo list to create a structure that maintains the list of sequence numbers of active connection IDs, because without that, we cannot see if the RCID frame that we receive is a duplicate of a previously received frame.

include/quicly/sentmap.h
```diff
@@ -109,6 +114,12 @@ struct st_quicly_sent_t {
             int is_inflight;
             uint64_t generation;
         } new_token;
+        struct {
+            struct st_quicly_pending_new_cid_t *new_cid;
+        } new_connection_id;
+        struct {
+            struct st_quicly_pending_retire_cid_t *retire_cid;
```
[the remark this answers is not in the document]
hfujita: I totally like the idea to reduce (or cap) the amount of resource to track state. Just one question here:
>  If the list is full, we simply drop the CID without ever sending an RCID frame for it.

What will eventually happen to this CID? Do we ever get a chance to re-send it by being triggered by some other event? Peer including the sequence in retire-prior-to of NCID is one example, but what happens if retire-prior-to stays lower than the sequence number in question?

lib/quicly.c
```diff
@@ -284,9 +301,28 @@ struct st_quicly_conn_t {
...
          */
         quicly_cc_t cc;
+        /**
+         * pending NEW_CONNECTION_ID
+         */
+        struct {
+            /**
+             * maximum number of path ID issued so far
+             */
+            uint32_t max_path_id;
```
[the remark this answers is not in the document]
hfujita: 👍 Actually I had a same concern but couldn't instantly come up with a better name. Thanks for your suggestions! I'd go with "last_path_id" (which reflects the current logic in the code).

lib/quicly.c
```diff
@@ -3345,6 +3494,53 @@ static int send_connection_close(quicly_conn_t *conn, quicly_send_context_t *s)
...
+    quicly_cid_t cid;
+    uint8_t stateless_reset_token[QUICLY_STATELESS_RESET_TOKEN_LEN];
+    quicly_sent_t *sent;
+    quicly_cid_encryptor_t *cid_encryptor = conn->super.ctx->cid_encryptor;
+    uint64_t retire_prior_to = 0; /* TODO */
+
+    plain_cid.path_id = new_cid->path_id;
+
+    assert(cid_encryptor != NULL);
+    cid_encryptor->encrypt_cid(cid_encryptor, &cid, stateless_reset_token, &plain_cid);
```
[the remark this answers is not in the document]
hfujita: You're totally right. I somehow (incorrectly) assumed that `encrypt_cid` was purely functional. Thank you for catching it.

include/quicly/sentmap.h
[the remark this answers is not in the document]
kazuho: > What will eventually happen to this CID?

From the viewpoint of the peer, it would look like it has not been retired. And I think that is fine. It would be the consequence of consuming CIDs too fast.

If that happens repeatedly on a single connection, the connection would eventually die. But it would be unlikely, as an endpoint should not be expecting the peer to issue CIDs at a high rate anyways; see https://quicwg.org/base-drafts/draft-ietf-quic-transport.html#section-5.1.1-5.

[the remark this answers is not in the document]
hfujita: Thanks for clarification. I'll go with your idea here then.

lib/quicly.c
```diff
@@ -284,9 +301,28 @@ struct st_quicly_conn_t {
...
+        struct {
+            /**
+             * maximum number of path ID issued so far
+             */
+            uint32_t max_path_id;
+            /**
+             * list of struct st_quicly_new_cid_t
+             */
+            quicly_linklist_t head;
+        } new_cid;
```
hfujita: @kazuho Should these (`new_cid` and `retire_cid` structures here) under `pending` instead of `egress`?

include/quicly/frame.h
```diff
@@ -551,13 +565,51 @@ inline int quicly_decode_streams_blocked_frame(const uint8_t **src, const uint8_
     return 0;
 }
 
+static inline size_t quicly_new_connection_id_frame_capacity(uint64_t sequence, uint64_t retire_prior_to, uint8_t cid_len)
```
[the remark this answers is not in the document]
hfujita: Done.

include/quicly/sentmap.h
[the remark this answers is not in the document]
hfujita: Done.

include/quicly/sentmap.h
```diff
@@ -109,6 +114,12 @@ struct st_quicly_sent_t {
             int is_inflight;
             uint64_t generation;
         } new_token;
+        struct {
+            struct st_quicly_pending_new_cid_t *new_cid;
```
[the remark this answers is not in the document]
hfujita: Done.

lib/quicly.c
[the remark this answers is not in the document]
hfujita: Done.

lib/quicly.c
```diff
@@ -284,9 +301,28 @@ struct st_quicly_conn_t {
...
+         */
+        struct {
+            /**
+             * maximum number of path ID issued so far
+             */
+            uint32_t max_path_id;
+            /**
+             * list of struct st_quicly_new_cid_t
+             */
+            quicly_linklist_t head;
```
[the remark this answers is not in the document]
hfujita: Done.

lib/quicly.c
```diff
@@ -284,9 +301,28 @@ struct st_quicly_conn_t {
...
+        } new_cid;
+        /**
+         * pending RETIRE_CONNECTION_ID frames to be sent, as a response to retire prior to in NEW_CONNECTION_ID
+         */
+        struct {
+            quicly_linklist_t head;
+        } retire_cid;
```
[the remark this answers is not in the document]
hfujita: Done.

lib/quicly.c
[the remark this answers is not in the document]
hfujita: Done.

lib/quicly.c
hfujita: Done for the first part (when obtaining 1-RTT write key). The second part is going to be on another round of PR.

include/quicly.h
hfujita: 👍 Done.

include/quicly.h
hfujita: Sounds good. Done.

include/quicly.h
hfujita: Yes, it's coming in the next round of PR. What I'm thinking right now is to have an array like `int retired_cids[MAX_PATH_ID];` to indicate which one has been retired.

hfujita:
@kazuho I think I have addressed all of your comments. Please take a look when you get a chance.

include/quicly/sentmap.h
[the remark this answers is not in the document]
hfujita: In fact I totally missed `quicly_get_first_timeout` in the initial version of the PR. Now it has been added. Thanks for pointing it out!

hfujita:
For some reason I currently don't see a link to the CI result, but here it is https://travis-ci.org/github/h2o/quicly/builds/668976532 and the result was all green.

lib/quicly.c
```diff
@@ -175,6 +185,31 @@ struct st_quicly_application_space_t {
...
+    /**
+     * this entry has been sent and is waiting for ACK (or timeout)
+     */
+    QUICLY_PENDING_NEW_CID_STATE_INFLIGHT,
+};
+
+/**
+ * records information for sending NEW_CONNECTION_ID frame
+ */
+struct st_quicly_pending_new_cid_t {
```
kazuho, line 206: Regarding the name of the structure and the necessary states, please refer to my comments below.

include/quicly/sentmap.h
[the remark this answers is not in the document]
kazuho: Excuse me, let me retract my comment. I do not think we can use a pointer. We need to retain the sequence number. Please see the discussion below regarding the state transitions.

lib/quicly.c
```diff
@@ -4367,14 +4674,155 @@ static int handle_ping_frame(quicly_conn_t *conn, struct st_quicly_handle_payloa
...
 {
+    int ret;
+    quicly_retire_connection_id_frame_t frame;
+
+    if ((ret = quicly_decode_retire_connection_id_frame(&state->src, state->end, &frame)) != 0)
+        return ret;
+
+    /* TODO: return PROTOCOL_VIOLATION if sequence is associated with a DCID of this packet */
+
+    /* TODO: implement retirement logic */
```
kazuho, line 4823: I now think we need to implement this before doing the next round of review. As stated above, this is inherently tied to how we handle the delivery of NCID frames.

include/quicly.h
```diff
@@ -471,16 +489,31 @@ struct _st_quicly_conn_public_t {
         /**
          * stateless reset token corresponding to the CID
          */
-        struct {
+        struct st_quicly_stateless_reset_t {
```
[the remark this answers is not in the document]
hfujita: 👍 Good catch. I just forgot to remove the name when reverting the generalization patch.

include/quicly/frame.h
```diff
@@ -563,13 +577,50 @@ inline int quicly_decode_streams_blocked_frame(const uint8_t **src, const uint8_
     return 0;
 }
 
+inline size_t quicly_new_connection_id_frame_capacity(uint64_t sequence, uint64_t retire_prior_to, uint8_t cid_len)
+{
+    return 1 + quicly_encodev_capacity(sequence) + quicly_encodev_capacity(retire_prior_to) + cid_len +
```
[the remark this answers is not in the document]
hfujita: 😲 Wow... thank you for catching this. This would have caused a nasty memory corruption bug in future otherwise.

lib/quicly.c
```diff
@@ -78,6 +79,15 @@
...
 #define QUICLY_NUM_ACK_BLOCKS_TO_INDUCE_ACKACK 8
+/**
+ * up to how many RETIRE_CONNECTION_IDs to keep for retransmission
+ */
+#define QUICLY_RETIRE_CONNECTION_ID_LIMIT (QUICLY_LOCAL_ACTIVE_CONNECTION_ID_LIMIT * 2)
+/**
+ * indicates there's nothing to send in this retire CID slot
+ * see quicly_conn_t::egress::retire_cid
+ */
+#define QUICLY_PENDING_RETIRE_CID_EMPTY UINT64_MAX
```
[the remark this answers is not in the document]
hfujita: Ah okay, I thought `UINT64_MAX` was too rude 😆 I'll rename it to `UINT64_MAX`.

include/quicly.h
hfujita: Makes sense. I'll go with `uint64_t` as I also hate the maintenance burden with `uint8_t`.

include/quicly/sentmap.h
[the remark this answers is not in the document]
hfujita: I see. The reason is ... when the ACL arrives there is a chance that the corresponding CID has already been retired through RCID frame, right?

lib/quicly.c
```diff
@@ -284,9 +319,43 @@ struct st_quicly_conn_t {
...
+         */
+        struct {
+            /**
+             * maximum number of path ID issued so far
+             */
+            uint32_t last_path_id;
+            /**
+             * list of struct st_quicly_pending_new_cid_t
+             */
+            struct st_quicly_pending_new_cid_t *cids;
```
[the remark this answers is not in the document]
hfujita: Hmm, are you introducing a new quicly-internal limit `N`, which is basically independent from `MAX_PATH_ID` or `peer.tp.active_connection_id_limit` (*)?

If that's the case I think it makes sense. In case the peer announced `peer.tp.active_connection_id_limit = 255`, it sounds too much to issue 254 CIDs at once (which is what my code currently does.)

(*) given `N <= min(MAX_PATH_ID, peer.tp.active_connection_id_limit)` is satisfied

lib/quicly.c
hfujita: Okay I think I got the idea. Let me work on this.

In the meantime, do you have any opinion on where to place `struct { ... } retire_cid`?

include/quicly/sentmap.h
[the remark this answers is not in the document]
kazuho: Correct.

lib/quicly.c
kazuho: How about making it a member of `egress`? I think we might want to merge `penging` into `egress`.

lib/quicly.c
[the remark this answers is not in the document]
kazuho: What you have stated is exactly what I meant.

And the limit can be equal to local.tp.active_connection_id_limit. This is because client-issued CIDs and server-issued CIDs are typically used in pair.

lib/quicly.c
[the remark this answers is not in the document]
hfujita: Well... let me take it back. Actually the current usage was something equivalent to `next_path_id`. I'll rename it again to make it clear.

lib/quicly.c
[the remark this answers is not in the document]
hfujita: Coming back on this topic. Given `cid_encryptor` may change over time, I wonder if we could have a scenario like this:
1. server issues multiple CIDs upon connection establishment
2. cid_encryptor changes in the server process
3. client switches to one of the CIDs issued in step 1
4. server tries to decrypt the CID but since cid_encryptor is different from what it had when the CID was issued (step 1), it cannot correctly obtain `cid_plaintext_t`, and things like `quicly_is_destination` don't work as expected.

Is this something that could actually happen, and if so, is it something to worry about? Maybe the probability of hitting such a case is low enough?

hfujita:
@kazuho I think I have addressed all of your comments so far. I've also implemented the RETIRE_CONNECTION_ID handler (`handle_retire_connection_id_frame`) as you suggested. Please take another look.

As a simple test, in my local environment I forcefully set "retire prior to field" to 1 when sending a NEW_CONNECTION_ID and manually verified that RETIRE_CONNECTION_ID frame was actually sent/received and the client switched to another CID when talking to the server.

lib/quicly.c
```diff
@@ -309,6 +343,19 @@ struct st_quicly_conn_t {
...
+         */
+        struct {
+            /**
+             * sequence numbers to ask for retirement -- UINT64_MAX indicates empty slot
+             */
+            uint64_t sequences[QUICLY_RETIRE_CONNECTION_ID_LIMIT];
+            /**
+             * # of active elements in sequences
+             */
+            uint64_t num_pending;
```
kazuho, line 357: Instead of having `num_pending`, I kind of wonder if it would be possible to use a bit of `pending_flows` to indicate if there's more than one entry in `sequence`?

That would reduce the number of CPU cycles spent in `quicly_get_first_timeout`, and also free us from the burden of mainitaning `sequences` and `num_pending` in sync.

`sequences` can be a list of valid sequence numbers that is terminated by `UINT64_MAX`. When adding a CID to the list, you would append a CID, then, when sending RCID frames, you would pop one entry from the beginning of the list.

lib/quicly.c
```diff
@@ -2321,6 +2537,10 @@ int64_t quicly_get_first_timeout(quicly_conn_t *conn)
             return 0;
         if (scheduler_can_send(conn))
             return 0;
+        if (conn->issued_cid.num_pending > 0)
+            return 0;
+        if (conn->egress.retire_cid.num_pending > 0)
+            return 0;
```
kazuho, line 2543: Please see my comments above.

lib/quicly.c
```diff
@@ -4375,15 +4681,184 @@ static int handle_ping_frame(quicly_conn_t *conn, struct st_quicly_handle_payloa
...
+    QUICLY_PROBE(NEW_CONNECTION_ID_RECEIVE, conn, probe_now(), frame.sequence, frame.retire_prior_to);
+
+    if (frame.sequence < conn->super.peer.max_retire_prior_to) {
+        /* An endpoint that receives a NEW_CONNECTION_ID frame with a sequence number smaller than the Retire Prior To
+         * field of a previously received NEW_CONNECTION_ID frame MUST send a corresponding RETIRE_CONNECTION_ID frame
+         * that retires the newly received connection ID, unless it has already done so for that sequence number. (19.15)
+         * TODO: "unless ..." part may not be properly addressed here (we may already have sent the RCID frame for this
+         * sequence) */
+        if (schedule_retire_connection_id(conn, frame.sequence) != 0)
+            return QUICLY_TRANSPORT_ERROR_INTERNAL;
```
kazuho, line 4731: What does the return code of `schedule_retire_connection_id` mean? If it is an error code, then I think we should be returning the value returned by the function.

As an alternative, I think we can change the return type of `schedule_retire_connection_id` to void?

lib/quicly.c
```diff
@@ -4375,15 +4681,184 @@ static int handle_ping_frame(quicly_conn_t *conn, struct st_quicly_handle_payloa
...
+        if (frame.retire_prior_to > conn->super.peer.cid_sequence) {
+            if (schedule_retire_connection_id(conn, conn->super.peer.cid_sequence) != 0) {
+                /* TODO: if the scheduling fails (likely due to malloc failure),
+                 * it's possible that we could continue without disrupting the connection
+                 * -- just silently ignore retire_prior_to and the peer will eventually
+                 * retransmit the same information.
+                 * However given that malloc failure indicates catastrophic events are going
+                 * on in the system, probably it's reasonable to drop the connection by
+                 * returning the error. */
+                return QUICLY_TRANSPORT_ERROR_INTERNAL;
```
kazuho, line 4750: Please see my comment above.

lib/quicly.c
```diff
@@ -4375,15 +4681,184 @@ static int handle_ping_frame(quicly_conn_t *conn, struct st_quicly_handle_payloa
...
+            conn->super.peer.cid_sequence = frame.sequence;
+            return 0;
+        }
+        /* replace with the picked one */
+        spare_cid->is_active = 0;
+        conn->super.peer.cid = spare_cid->cid;
+        memcpy(conn->super.peer.stateless_reset._buf, spare_cid->stateless_reset_token, QUICLY_STATELESS_RESET_TOKEN_LEN);
+        conn->super.peer.stateless_reset.token = conn->super.peer.stateless_reset._buf;
+        conn->super.peer.cid_sequence = spare_cid->sequence;
+    }
```
kazuho, line 4783: The logic here has two ways of picking a new CID. I wonder if we can merge the two into one, if we change the logic of this function to at first register the value carried by the NCID frame then pick one from the spare list.

include/quicly/constants.h
```diff
@@ -36,6 +36,10 @@ extern "C" {
 #define QUICLY_LOCAL_MAX_ACK_DELAY 25   /* milliseconds */
 #define QUICLY_DEFAULT_ACK_DELAY_EXPONENT 3
 #define QUICLY_LOCAL_ACK_DELAY_EXPONENT 10
+#define QUICLY_DEFAULT_ACTIVE_CONNECTION_ID_LIMIT 2 /* If this transport parameter is absent, a default of 2 is assumed. (18.2) */
+/** how many CIDs is quicly willing to manage at the same time? */
+#define QUICLY_LOCAL_ACTIVE_CONNECTION_ID_LIMIT 4
```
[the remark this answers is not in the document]
hfujita: Yes as you suspected it serves two purposes. I'll update the comment to clarify it.

lib/quicly.c
hfujita: 👍 Reducing `if` branch in a critical path sounds too attractive to ignore :) Let me try to see if this works.

lib/quicly.c
```diff
@@ -340,6 +387,23 @@ struct st_quicly_conn_t {
...
+        struct st_quicly_issued_cid_t cids[QUICLY_LOCAL_ACTIVE_CONNECTION_ID_LIMIT];
+        /**
+         * # of elements in cids whose state is PENDING
+         */
+        uint64_t num_pending;
+        /**
+         * # of retired CIDs
+         */
+        uint64_t num_retired;
+    } issued_cid;
```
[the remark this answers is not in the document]
hfujita: Here's what I understood from your comments (beside using `pending_flows`):
Instead of trying to find an empty (i.e. `IDLE`) slot each time, introduce a simple principle that elements `[0, ..., N-1]` shall always be non-idle. Any `IDLE` entry between `[0, N-1]` shall be converted to `PENDING` so that corresponding NCID frame will be sent in the next send round.

If I understood correctly, I think that's worth pursuing. Let me try to implement it.

lib/quicly.c
```diff
@@ -4375,15 +4681,184 @@ static int handle_ping_frame(quicly_conn_t *conn, struct st_quicly_handle_payloa
...
+    for (int i = 0; i < QUICLY_LOCAL_ACTIVE_CONNECTION_ID_LIMIT; i++) {
+        struct st_quicly_issued_cid_t *c = &conn->issued_cid.cids[i];
+
+        if (c->state == QUICLY_ISSUED_CID_STATE_IDLE || c->sequence != frame.sequence)
+            continue;
+
+        if (c->state == QUICLY_ISSUED_CID_STATE_PENDING)
+            conn->issued_cid.num_pending--;
+        c->state = QUICLY_ISSUED_CID_STATE_IDLE;
+        if (++conn->issued_cid.num_retired >= QUICLY_MAX_PATH_ID - 1) {
```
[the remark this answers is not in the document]
hfujita: I think you are right. Removed this check in the latest version.

lib/quicly.c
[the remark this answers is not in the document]
hfujita: Implemented the new logic based on my understanding above.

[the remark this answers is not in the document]
hfujita: One thing maybe worth discussing is the way I handle the bit field. Currently I use one bit to represent if there is any pending NCID or RCID. I did not introduce separate bits for NCID and RCID, because it would need to extend `pending_flows` to `uint16_t`. Also, I thought NCID/RCID were rarely sent, so it would make sense to consolidate then into one bit.

Let me know what you think.

lib/quicly.c
hfujita: Good catch. `schedule_retire_connection_id` used to fail (had a malloc) but now does not. Changed the function type to `void` to simplify the code.

lib/quicly.c
hfujita: 👍 Done.

hfujita:
@kazuho I believe I have addressed all of your latest comments. Please take another look, thanks!

lib/quicly.c
```diff
@@ -4444,15 +4724,177 @@ static int handle_ping_frame(quicly_conn_t *conn, struct st_quicly_handle_payloa
     return 0;
 }
 
+static struct st_quicly_spare_cid_t *pick_spare_cid(quicly_conn_t *conn)
+{
+    struct st_quicly_spare_cid_t *cid = NULL;
+    /* find the CID with smallest sequence number */
+    for (int i = 0; i < QUICLY_LOCAL_ACTIVE_CONNECTION_ID_LIMIT - 1; i++) {
```
kazuho, line 4731: size_t please.

lib/quicly.c
```diff
@@ -4444,15 +4724,177 @@ static int handle_ping_frame(quicly_conn_t *conn, struct st_quicly_handle_payloa
...
+    spare_cid->sequence = frame->sequence;
+    set_cid(&spare_cid->cid, frame->cid);
+    memcpy(spare_cid->stateless_reset_token, frame->stateless_reset_token, QUICLY_STATELESS_RESET_TOKEN_LEN);
+    spare_cid->is_active = 1;
+}
+
+/**
+ * compares CIDs, returns zero if identical
+ */
+static int cmp_cid(const quicly_cid_t *l, ptls_iovec_t r)
```
kazuho, line 4758: Could this be changed to a boolean function that returns true if the two are equal, e.g., `cid_is_equal`?

"cmp" is ambigous. It might imply that it is return a ternary value that can be used for sorting the values being provided, much like `memcmp`. But this function is not designed for such purpose, and IMO there is no need to.

lib/quicly.c
```diff
@@ -4444,15 +4724,177 @@ static int handle_ping_frame(quicly_conn_t *conn, struct st_quicly_handle_payloa
...
+     * one new CID.
+     */
+    if (frame.retire_prior_to > conn->super.peer.largest_retire_prior_to) {
+        if (frame.retire_prior_to > conn->super.peer.cid_sequence) {
+            schedule_retire_connection_id(conn, conn->super.peer.cid_sequence);
+            need_to_replace_cid = 1;
+        }
+
+        /* retire CIDs in the spare pool */
+        for (int i = 0; i < QUICLY_LOCAL_ACTIVE_CONNECTION_ID_LIMIT - 1; i++) {
```
kazuho, line 4800: Instead of `QUICLY_LOCAL_ACTIVE_CONNECTION_ID_LIMIT - 1`, could we use `sizeof(conn->super.peer.spare_cids) / sizeof(conn->super.peer.spare_cids[0])`?

They are identical, but IMO the latter is better, as this for block would only depend on the definition of `spare_cids`, rather than depending on the definition of `spare_cids` and `QUICLY_LOCAL_ACTIVE_CONNECTION_ID_LIMIT`.

lib/quicly.c
```diff
@@ -4444,15 +4724,177 @@ static int handle_ping_frame(quicly_conn_t *conn, struct st_quicly_handle_payloa
...
+                }
+            }
+            if (spare_cid->sequence == frame.sequence && cmp_cid(&spare_cid->cid, frame.cid) != 0)
+                return QUICLY_TRANSPORT_ERROR_PROTOCOL_VIOLATION;
+
+            continue;
+        }
+
+        store_spare_cid(spare_cid, &frame);
+        was_stored = 1;
```
kazuho, line 4838: Now I realize that we have a problem in the design.

Consider the case where we have received NCID(seq=1), used that CID, sent RCID(seq=1), and received an ACK for RCID(seq=1). After that, the peer might send NCID(seq=1. That is legal behavior, see https://quicwg.org/base-drafts/draft-ietf-quic-transport.html#section-13.3-5).

In such case, we'd be storing CID<sub>1</sub> and might end up using it, which we shouldn't be.

IMO, what we should be doing is something like the following:
* When we establish a connection, allocate slots in `spare_cids` for seq=1, seq=2, ..., and also retain the maximum sequence number that the peer is allowed to send at this moment.
* When receiving a NEW_CONNECTION_ID frame, check NCID.sequence against that maximum. If it is not above that maximum, check if there is an entry within the `spare_cids` array with given sequence number. If there is, store the given CID there. If there is no corresponding entry within the array, we know that we've already retired that CID.
* When we retire a CID (or handle RTP), we update `space_cid` and the allowed maximum as appropriate.

lib/quicly.c
```diff
@@ -3619,6 +3872,28 @@ static int do_send(quicly_conn_t *conn, quicly_send_context_t *s)
...
+                    }
+                    /* send RETIRE_CONNECTION_ID */
+                    for (int i = 0; i < QUICLY_RETIRE_CONNECTION_ID_LIMIT; i++) {
+                        uint64_t sequence = conn->egress.retire_cid.sequences[i];
+                        if (sequence == UINT64_MAX)
+                            continue; /* empty slot */
+                        if ((ret = send_retire_connection_id(conn, s, sequence)) != 0)
+                            goto Exit;
+                        conn->egress.retire_cid.sequences[i] = UINT64_MAX;
+                    }
```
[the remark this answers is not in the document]
hfujita: I'm a little confused. So sounds like your first priority is to keep consistency between NCID and RCID handling. Given such a consistency is maintained, you would prefer making the arrays like FIFO. It's trivial for RCID but may not be for NCID. If it's too hard to make NCID a FIFO, then you would stick to the current approach (non-FIFO) for consistency.

Did I understand your intention correctly?

lib/quicly.c
```diff
@@ -2335,6 +2487,55 @@ static int on_ack_new_token(quicly_conn_t *conn, const quicly_sent_packet_t *pac
...
+    return 0;
+}
+
+static int on_ack_retire_connection_id(quicly_conn_t *conn, const quicly_sent_packet_t *packet, quicly_sent_t *sent,
+                                       quicly_sentmap_event_t event)
+{
+    uint64_t sequence = sent->data.retire_connection_id.sequence;
+
+    if (event == QUICLY_SENTMAP_EVENT_EXPIRED)
+        return 0;
```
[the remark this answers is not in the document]
hfujita: 👍 Good catch. Will remove it.

lib/quicly.c
```diff
@@ -3418,6 +3619,45 @@ static int send_connection_close(quicly_conn_t *conn, quicly_send_context_t *s)
...
+        conn, s, quicly_new_connection_id_frame_capacity(new_cid->sequence, retire_prior_to, new_cid->cid.len), &sent,
+        on_ack_new_connection_id);
+    if (ret != 0)
+        return ret;
+    sent->data.new_connection_id.sequence = new_cid->sequence;
+
+    s->dst = quicly_encode_new_connection_id_frame(s->dst, new_cid->sequence, retire_prior_to, new_cid->cid.cid, new_cid->cid.len,
+                                                   new_cid->stateless_reset_token);
+
+    QUICLY_PROBE(NEW_CONNECTION_ID_SEND, conn, probe_now(), new_cid->sequence, retire_prior_to);
```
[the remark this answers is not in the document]
hfujita: Yeah... until now I didn't realize I could do it. Just added them using `QUICLY_PROBE_HEXDUMP`.

lib/quicly.c
hfujita: 👍

lib/quicly.c
```diff
@@ -774,6 +839,64 @@ static int schedule_path_challenge(quicly_conn_t *conn, int is_response, const u
     return 0;
 }
 
+/**
+ * calculate how many CIDs we provide to the peer
+ */
+static uint64_t issued_cid_capacity(const quicly_conn_t *conn)
```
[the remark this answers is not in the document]
hfujita: Sure. I was actually wondering whether it should be `size_t` or `uint64_t`. I'll go with `size_t` as you suggest.

As of now, I believe `prepare_pending_new_connection_id` is safe because its caller is always checking if `path_id < MAX_PATH_ID` (note: `path_id == MAX_PATH_ID` when `cid_encryptor == NULL`). But anyway I think it's a good idea to guard against future mistakes in multiple layers. I'll add a check in `issued_cid_capacity`.

lib/quicly.c
hfujita: Makes sense. Alternatively I could define a new constant like `QUICLY_SPARE_CONNECTION_ID_LIMIT` and use it for both definition and usage of `spare_cids[]`, as I personally feel writing down `sizeof(conn->super.peer.spare_cids) / sizeof(conn->super.peer.spare_cids[0])` is a bit annoying.

lib/quicly.c
hfujita: You're right. Thanks for noticing this.

hfujita:
Thanks for the review. I agree that having unit tests are useful/necessary. Let me take a look into it.

> Then, we could write unit tests without dealing with the connection object.

I will need to further dig into this, but at the moment I'm not sure if I can easily take out the dependency to the connection object in unit test path. I'm thinking about allocating & initializing some essential fields of a connection object and pass it for the unit test, if detaching the connection object turns out to be too hard.

kazuho:
> I agree that having unit tests are useful/necessary. Let me take a look into it.
>
>> Then, we could write unit tests without dealing with the connection object.
> 
> I will need to further dig into this, but at the moment I'm not sure if I can easily take out the dependency to the connection object in unit test path. I'm thinking about allocating & initializing some essential fields of a connection object and pass it for the unit test, if detaching the connection object turns out to be too hard.

I understand the concern. I do not want to ask many things at once, as that raises the probability of getting us derailed. Maybe one of the ways moving forward would be to first finish working on this PR, then refactor it so that unit tests can be added.

lib/quicly.c
hfujita: I have no doubt we must deal with scenario like you mentioned above.

Having said that, I'm just wondering under what scenario this (NCID with seq=1 being resent) would happen. Some examples I could think are: late arrival of duplicated packet and malicious/buggy peer. Would you mind elaborating more on this?

lib/quicly.c
kazuho: How about creating `quicly_countof` or something alike https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/countof-macro?view=vs-2019?

I agree that using `sizeof` directly is tedious, and IMO that is because we need to write the same expression twice. I think we'd be fine if we could change this to something like `i < quicly_countof(conn->super.peer.spare_cids)`.

I'm not very much in favor of defining another constant that represents the number of elements of the array, because we need to remember that that constant is used as the number of elements of that array.

lib/quicly.c
kazuho: As pointed out earlier, the specification allows a QUIC endpoint to retransmit the entire payload of a packet being sent previously. Keeping that in mind, let's consider the following case, where the egress path is very lossy:
1. endpoint repeatedly receives NCID(CID1), sends ACKs, but the ACKs get lost
2. endpoint uses CID1 to probe a new path
3. path probing fails
4. endpoint sends RCID(CID1)
5. endpoint receives ACK for RCID(CID1)

The point is that there is no guarantee that one of the ACKs sent in step 1 reaches the peer before the peer sends ACK for RCID(CID1). We might not hit this situation in practice, though I'm not sure if that is guaranteed. I'd note that path probing might fail almost immediately in certain cases, for example when the interface that was used for sending the probe becomes unusable.

I think we can consider a similar situation when RTP is involved.

lib/quicly.c
hfujita: Makes sense. Now that I see #325, I'll use it once it gets merged.

lib/quicly.c
hfujita: 👍 Thanks for a good example!

hfujita: > When receiving a NEW_CONNECTION_ID frame, check NCID.sequence against that maximum. If it is not above that maximum, check if there is an entry within the spare_cids array with given sequence number. If there is, store the given CID there. If there is no corresponding entry within the array, we know that we've already retired that CID.

Regarding detection of retired CID, I think quicly can have a principle that it would use CID in  the order of sequence number (which we currently do). That way the retired CID should have a sequence number smaller than the one currently in use for communication (`conn->super.peer.cid_sequence` in this PR).

hfujita: So the detection would look like
```
is_already_retired = frame.sequence < conn->super.peer.cid_sequence;
```
Let me know if you think of any potential flaw in this assumption.

lib/quicly.c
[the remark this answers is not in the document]
kazuho: Yes, what you have summarized reflects my (very weak) thoughts.

lib/quicly.c
hfujita: This doesn't prove the entire correctness of my algorithm above but as a note to myself here is what will happen when packet reordering happens. Let's say I'm using CID0 and the peer hasn't issued other CIDs yet. Then the peer issues CID1 and CID2:
```
CID1 (rpt=1)
CID2 (rpt=1)
```
but for whatever reason two frames are delivered out-of-order. So my host would see NCID frames in this order:
```
CID2 (rpt=1)
CID1 (rpt=1)
```
I handle CID2 first, and since RPT=1 is specified, CID0 is retired and CID2 becomes the current CID for communication. After that CID1 arrives, but since CID2 is already in use, CID1 is determined as "already retired" according to the above algorithm, so RCID for CID1 is sent.

This algorithm wastes some CIDs in cases like this but I would assume the probability of hitting such a case would be quite low, so I think it's still justifiable given its simplicity.

lib/quicly.c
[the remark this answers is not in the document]
hfujita: BTW can you explain why would you prefer FIFO to begin with? I think it makes sense in general... but for retransmissions, don't we want to prioritize retransmissions over first time transmissions?

lib/quicly.c
kazuho: Though nothing forbids us from using CIDs in the order of their sequence numbers, I am not sure if that is always going to work well.

As an example, https://quicwg.org/base-drafts/draft-ietf-quic-transport.html#section-9.1-2 suggests that the peer might send a probe with a NCID frame to make sure that an endpoint 
has a spare CID. But the specification does not say that the peer should send all unacknowledged NCID frames. My understanding is that this is because the specification assumes that endpoints would be capable of using CIDs in any order.

hfujita: Hmm, okay, then maybe we should not stick to the assumption that CIDs are sent and used in the order of sequence number.

hfujita: Updated the branch. I believe what I have now basically follows your algorithm.

hfujita:
@kazuho Finally updated the PR. I believe I have addressed your latest comments except for:
* make pending list of NCID/RCID FIFO
* add unit test for NCID/RCID handling

I think it is not too hard to convert NCID/RCID list into FIFO, but I thought I could address it later as an improvement.

Please take a look, and as always comments are much appreciated!

hfujita:
Apparently I missed an embarrassing test failure... now it's all green.

lib/quicly.c
```diff
@@ -4483,15 +4774,216 @@ static int handle_ping_frame(quicly_conn_t *conn, struct st_quicly_handle_payloa
...
+     *
+     * If the received sequence number exceeds largest_sequence_expected (after considering the CIDs to be retired via
+     * retire_prior_to), it is an ACTIVE_CONNECTION_LIMIT error.
+     * If the number is not found in the list of expected sequence numbers (conn->super.peer.spare_cids), it is considered to be
+     * already retired. */
+
+    uint64_t num_to_retire = frame.retire_prior_to > conn->super.peer.largest_retire_prior_to
+                                 ? frame.retire_prior_to - conn->super.peer.largest_retire_prior_to
+                                 : 0;
+    if (frame.sequence > conn->super.peer.largest_sequence_expected + num_to_retire) {
```
kazuho, line 4862: I'm not sure if this condition is true.

Consider the case where `largest_retire_prior_to` is zero and `frame.retire_prior_to` is 4. RHS of this comparison expression yields 4, and therefore we'd call it a protocol error when `frame.sequence` is any value above 4. Is that correct? I do not think so. As our local limit is 4, it is my understanding that sequence numbers up to 7 have to be considered valid.

Regarding how we should fix this issue, I think it might make sense to remove this check altogether. Instead, we could change the structure of this function to make changes in following steps:
1. Retire CIDs as instructed. At this point, the spare slots would be assigned a new sequence number, with the `is_active` being set to false. This entire logic can be coded in `schedule_retire_connection_id`.
2. Store the newly provided CID. At this point, we'd be looking for a slot that has the matching sequence number. If we could not find a slot with a matching sequence number, we know that the peer has sent something invalid.

kazuho: PS. To paraphrase, there are two operations that modify "the list of CIDs provided by the peer":
* when a potentially new CID is given
* when a CID is retired (regardless of the reason; e.g., us switching to a different address, peer using NCID.RPT)
Having two functions dedicated for each operation might make the code cleaner.

lib/quicly.c
```diff
@@ -4483,15 +4774,216 @@ static int handle_ping_frame(quicly_conn_t *conn, struct st_quicly_handle_payloa
...
+    spare_cid->sequence = frame->sequence;
+    set_cid(&spare_cid->cid, frame->cid);
+    memcpy(spare_cid->stateless_reset_token, frame->stateless_reset_token, QUICLY_STATELESS_RESET_TOKEN_LEN);
+    spare_cid->is_active = 1;
+}
+
+/**
+ * compares CIDs, returns zero if identical
+ */
+static int cid_is_equal(const quicly_cid_t *l, ptls_iovec_t r)
```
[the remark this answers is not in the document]
hfujita: Ouch. Not sure why I missed this. Will fix it.

lib/quicly.c
```diff
@@ -4483,15 +4774,216 @@ static int handle_ping_frame(quicly_conn_t *conn, struct st_quicly_handle_payloa
...
+ * @param sequence known sequence number
+ * @param cid known CID
+ * @param stateless_reset_token known stateless reset token
+ * @param frame newly received NEW_CONNECTION_ID frame
+ * @param retcode return code that the caller will return to its caller
+ * @return 0 if okay (the caller can continue), non-zero otherwise (the caller must terminate frame rocessing and return *retcode)
+ */
+static int verify_new_cid(uint64_t sequence, const quicly_cid_t *cid,
+                          const uint8_t stateless_reset_token[QUICLY_STATELESS_RESET_TOKEN_LEN],
+                          const quicly_new_connection_id_frame_t *frame, int *retcode)
```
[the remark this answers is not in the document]
hfujita: 👍 on renaming arguments.

For the return code, it's not that trivial. In case of valid but duplicated frame, this function has to tell the caller "it's okay but stop execution". In other words, simply returning zero doesn't distinguish "okay and continue" and "okay but return immediately".

lib/quicly.c
hfujita: > Consider the case where largest_retire_prior_to is zero and frame.retire_prior_to is 4. RHS of this comparison expression yields 4, and therefore we'd call it a protocol error when frame.sequence is any value above 4. Is that correct? I do not think so. As our local limit is 4, it is my understanding that sequence numbers up to 7 have to be considered valid.

Hmm I'm not sure if we are on the same page on how this works.

Let's say the host is announcing `active_connection_id_limit = 4` and got 4 CIDs (including the initial one at the connection establishment). At this moment `largest_sequence_expected == 3`. And as in your example, let's say largest_retire_prior_to is zero and frame.retire_prior_to is 4. Then this expression
```
conn->super.peer.largest_sequence_expected + num_to_retire
```
gives 3+4=7. Thus `frame.sequence == 7` will be accepted as a valid number.

hfujita: As we discussed offline, I'm more than happy to refactor it so that we can achieve the same goal with smaller LOC or much simpler code. So let me understand your idea more thoroughly.

As you pointed out probably I can get rid of the comparison `if (frame.sequence > conn->super.peer.largest_sequence_expected + num_to_retire)`. The only reason I put it upfront was that I was reluctant to modify the state (e.g. `spare_cids`) if the frame is going to be determined invalid, but since we'll close the connection in case of invalid frame and dispose all of per-connection states, maybe we don't have to worry about it.

> Retire CIDs as instructed. At this point, the spare slots would be assigned a new sequence number, with the is_active being set to false. This entire logic can be coded in schedule_retire_connection_id.

To be precise I'm skeptical about it. Currently `schedule_retire_connection_id` is used in many other places such as retiring the current CID for communication and rescheduling RCID after packet loss, in which cases there exists no corresponding `spare_cid_t` entry.

> To paraphrase, there are two operations that modify "the list of CIDs provided by the peer": ... (snip) ... Having two functions dedicated for each operation might make the code cleaner.

I think the "new CID is given" part is `store_spare_cid` function. I agree that it would be better to have a counterpart for deactivating the spare (i.e. setting `is_active` to false and set the next expected sequence number). Is that what you are suggesting?

kazuho: > The only reason I put it upfront was that I was reluctant to modify the state (e.g. `spare_cids`) if the frame is going to be determined invalid, but since we'll close the connection in case of invalid frame and dispose all of per-connection states, maybe we don't have to worry about it.

:+1:

>> Retire CIDs as instructed. At this point, the spare slots would be assigned a new sequence number, with the is_active being set to false. This entire logic can be coded in schedule_retire_connection_id.
> 
> To be precise I'm skeptical about it. Currently `schedule_retire_connection_id` is used in many other places such as retiring the current CID for communication and rescheduling RCID after packet loss, in which cases there exists no corresponding `spare_cid_t` entry.

:+1:

>> To paraphrase, there are two operations that modify "the list of CIDs provided by the peer": ... (snip) ... Having two functions dedicated for each operation might make the code cleaner.
> 
> I think the "new CID is given" part is store_spare_cid function. I agree that it would be better to have a counterpart for deactivating the spare (i.e. setting is_active to false and set the next expected sequence number). Is that what you are suggesting?

Kind of. That matches what I'd expect the `unregister_cid` method to do in the example below. But let's step back for a moment.

My concern is that this is a complex stuff. That is due to the spec being complex, the people involved in the standardization (including me) are to blame.

I am very happy to see this PR having all the necessary code that implements the complex spec, thank you very much for your efforts. Now might be a good time to start refactoring our design to make sure that we have all things correctly laid out, and so that it could be maintained without the fear of breaking something.

Considering that, I think it would be a good idea to split frame processing and state processing. Also, if we could possibly create an object (or objects) that retains the states, it would be easier to write unit tests just regarding the state processing.

Regarding how an ideal code might look like, we might think of something like below.

```c++
/**
 * Retains active CIDs.
 */
struct ActiveCIDs {
    struct {
        uint64_t sequence;
        quicly_cid_t cid;
        uint8_t srt[...];
    } active_cids[4];
    /**
     * Initializez the object.
     */
    ActiveCIDs();
    /**
     * Registers a new CID.
     * Returns 0 if successful, otherwise a transport error.
     */
    int register_cid(uint64_t sequence, ptls_iovec_t cid, const void *srt);
    /**
     * Unregisters a CID, given the index of the CID within `active_cids`.
     * The caller sends an RCID frame after calling this function.
     */
    void unregister_cid(size_t cid_index);
}
```

This is just an example. This example assumes that we handle the current (in-flight) CID and the spares in the same way (we can, but we do not do that now).

Also, this object does not keep RPT by itself. That's because when the object is going to process RPT, it needs to signal retirement of multiple CIDs. However, that's going to require a callback, or having a way to pass a list of sequence numbers. We can do that, or we could retain the pending active CIDs _and pending _RCIDs_ in a single object. But as I was uncertain if that is a good design, I am providing here a simple example.

kazuho: PS. To be honest, if I were you, I would be worried about possibly breaking this PR if we are to do a refactor like this. And because I would be worried of that, I would create a new PR against this PR for that purpose, _if_ I would try to see if we can make the code simpler by refactoring the code.

lib/quicly.c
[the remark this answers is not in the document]
hfujita: Renaming done.

lib/quicly.c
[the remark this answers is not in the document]
hfujita: Done.

lib/quicly.c
hfujita: Thanks for providing ideas. Actually, to my embarrassment, it hasn't occurred to me that issued CIDs and received CIDs can be (and perhaps should be) handled in a common data type. Looks like a good idea. Let me dig into this direction and get back to you. Meanwhile I've pushed trivial fixes and cleanups.

lib/quicly.c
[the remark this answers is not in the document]
kazuho: Thank you for renaming the arguments. With the changes, I think it has become easier to read.

Re return code, reading your comment, I now realize that this function is used for two purposes: (1) verify if the provided values are correct, (2) signal to the caller that there's nothing to be done.

I think my concern now is that it's confusing, regarding the name and how it's used.

```c
if (verify_new_cid(conn->super.peer.cid_sequence, &conn->super.peer.cid, conn->super.peer.stateless_reset.token, &frame, &ret))
         return ret;
```

It's a bit difficult to understand that we are deviating from our ordinary way of doing `if ((ret = foo()) != 0) return ret`, and that in this particular case we might be returning zero.

How about dropping the (2)? I believe that can be done in the caller; i.e. `if (seq == frame.seq) return 0`.

Or, we could lift the implication that the function is merely verifying something, by changing the name to something like "check_new_cid" and do:
```c
if ((ret = check_new_cid(conn->super.peer.cid_sequence, &conn->super.peer.cid, &conn->super.peer.stateless_reset_token, &frame, &found)) != 0 || found)
    return ret;
```

Anyways, this is my slight preference, and it might be the case that you might not need to look into this now... we might change the way this function is implemented as we refactor the code.

lib/quicly.c
hfujita: > This example assumes that we handle the current (in-flight) CID and the spares in the same way (we can, but we do not do that now).

Now I realized that I misunderstood this comment. So this is about unifying the current CID and spare ones, not unifying structures for issued CIDs and received CIDs. . Within the received CIDs (current CID and spare ones), I'll explore a way to unify them, as it'll definitely simplify the logic.

Regarding unifying issued CIDs and received CIDs, though these two are quite similar but since they are used in totally disjoint contexts, for the moment I'd keep them separate.

hfujita: @kazuho Just submitted hfujita/quicly#1 based on your idea here. Please take a look and let me know what you think. If this direction looks okay, I'm thinking about migrating `issued_cid` state processing logic & `retire_cid` state processing logic into `cid.c` as well.

kazuho: Thank you for working on the refactor. I must say that I like it! It has become easier to read, and we have tests!

> If this direction looks okay, I'm thinking about migrating `issued_cid` state processing logic & `retire_cid` state processing logic into `cid.c` as well.

I agree that having them isolated too would help. Just curious, do you expect to have interaction between the three components (i.e. between `receive_cid`, `issued_cid`, and `retire_cid`), or is it just a coincidence that we'd have the three in one file?

hfujita: Thank you for a quick turnaround! At this moment I'm not sure if we're to see interaction between them. I just thought it'd be nice to separate code from quickly.c as much as reasonably possible.

kazuho: I see, thank you for the response.

To clarify, my weak preference goes to not having interaction between them. And if we want to be firm about that, I think we can use different files (pair of .h and .c files) for each of the three.

I do not think we need to be firm at this point, though. We can always split files at a later moment. I merely wanted to know if your idea of moving more logic into one file was based on the assumption that we'd be having interaction between the components stored in that one file.

hfujita:
@kazuho Finally completed the refactoring of the remaining two (`NEW_CONNECTION_ID` issuing and `RETIRE_CONNECTION_ID` issuing). Also, now these two logics implements the FIFO queuing as you suggested before. As a result of refactoring now they both have unit tests. Please take a look when you get a chance. Thanks!

hfujita:
So I'm seeing this build failure in macOS builds.
```
 /Users/travis/build/h2o/quicly/t/issued_cid.c:91:28: error: array size is negative
    for (size_t i = 0; i < PTLS_ELEMENTSOF(set->cids); i++) {
                           ^~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/travis/build/h2o/quicly/deps/picotls/include/picotls.h:49:29: note: expanded from macro 'PTLS_ELEMENTSOF'
#define PTLS_ELEMENTSOF(x) (PTLS_ASSERT_IS_ARRAY(x), sizeof(x) / sizeof((x)[0]))
                            ^~~~~~~~~~~~~~~~~~~~~~~
/Users/travis/build/h2o/quicly/deps/picotls/include/picotls.h:41:33: note: expanded from macro 'PTLS_ASSERT_IS_ARRAY'
#define PTLS_ASSERT_IS_ARRAY(a) PTLS_BUILD_ASSERT(__builtin_types_compatible_p(__typeof__(a[0])[], __typeof__(a)))
                                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/travis/build/h2o/quicly/deps/picotls/include/picotls.h:40:52: note: expanded from macro 'PTLS_BUILD_ASSERT'
#define PTLS_BUILD_ASSERT(cond) ((void)sizeof(char[2 * !!(!__builtin_constant_p(cond) || (cond)) - 1]))
                                                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```
Findings so far:
* Latest Xcode (LLVM 10.0) does NOT reproduce this failure
* Xcode 9.4 (Apple LLVM version 9.1.0 (clang-902.0.39.2)) DOES reproduce this failure
* On Ubuntu, clang-9 or older clang does NOT reproduce the failure
* In failing cases, files under `lib/`, which use the same `PTLS_ELEMENTSOF(set->cids)` expression, do not fail. Files under `t/` are failing, interestingly.

hfujita:
> Xcode 9.4 (Apple LLVM version 9.1.0 (clang-902.0.39.2)) DOES reproduce this failure

Turns out removing const qualifiers from the parameter type solves the problem. I'm tempted to say it's a compiler issue, otherwise it really doesn't make sense...
```
-static size_t num_pending(const quicly_issued_cid_set_t *set)
+static size_t num_pending(quicly_issued_cid_set_t *set)
```

hfujita:
In #339 I confirmed that using the latest Xcode (11.4) fixes the build issue. However there I saw this assertion failure, which I haven't been able to reproduce locally.
```
Assertion failed: (set->cids[i].state != QUICLY_ISSUED_CID_STATE_PENDING), function find_inflight, file /Users/travis/build/h2o/quicly/lib/issued_cid.c, line 108.
    not ok 2
    
    #   Failed test at t/e2e.t line 229.
    #                   ''
    #     doesn't match '(?^s:^hello world\n)'
    1..2
not ok 11 - 0-rtt-vs-hrr
```

kazuho:
Thank you for your continued work on this.

Regarding `PTLS_ELEMENTS_OF`, would you mind submitting a PR to picotls that makes `PTLS_ASSERT_IS_ARRAY_EXPR`a no-op when buggy clang versions are used? We already have something like https://github.com/h2o/picotls/blob/master/include/picotls.h#L49.

This macro is a helper to find buggy code at compile time, just having it enabled on _some_ of the environments is sufficient.

hfujita:
Latest CI failures in "loss" tests seem to exist in master too. I'll continue investigating the [assertion failure above](https://github.com/h2o/quicly/pull/306#issuecomment-626054112). I was able to reproduce it locally, with ~1/100 probability.

lib/retire_cid.c
```diff
@@ -0,0 +1,65 @@
...
+            /* already scheduled */
+            return;
+        }
+    }
+
+    assert(set->sequences[set->_num_pending] == UINT64_MAX);
+    set->sequences[set->_num_pending] = sequence;
+    set->_num_pending++;
+    if (set->_num_pending < PTLS_ELEMENTSOF(set->sequences))
+        set->sequences[set->_num_pending] = UINT64_MAX;
```
kazuho, line 51: Isn't this value guaranteed to contain `UINT64_MAX`?

Assuming that is the case, I think we can simply remove this if clause, or do an assert.

include/quicly/issued_cid.h
```diff
@@ -0,0 +1,155 @@
...
+static int quicly_issued_cid_is_empty(const quicly_issued_cid_set_t *set);
+static size_t quicly_issued_cid_get_capacity(const quicly_issued_cid_set_t *set);
+/**
+ * mark the first `num_sent` pending CIDs as INFLIGHT.
+ */
+void quicly_issued_cid_mark_inflight(quicly_issued_cid_set_t *set, size_t num_sent);
+/**
+ * mark the specified CID as DELIVERED.
+ *
+ * @return zero if successful, non-zero if the specified CID was not found.
```
kazuho, line 119: Our habit is to use 1 for success or 0 for failure when the function returns a boolean. Though I think this function can be `void`?

lib/issued_cid.c
```diff
@@ -0,0 +1,168 @@
...
+int quicly_issued_cid_mark_pending(quicly_issued_cid_set_t *set, uint64_t sequence)
+{
+    size_t i = find_inflight(set, sequence);
+    if (i == SIZE_MAX)
+        return 1;
+
+    do_mark_pending(set, i);
+
+    return 0;
+}
```
kazuho, line 139: I'm not sure if the logic in `quicly_issued_cid_mark_delivered` and `quicly_issued_cid_mark_pending` are correct.

When the endpoint receives a late ack, an entry marked as PENDING can transition to DELIVERED. Similarly, a packet (and therefore a frame) can be declared lost after another packet carrying the same information was being acked.

It might be a good idea to rename these functions to "_acked", "_lost" at the same time changing the logic to handle these cases. The reason I suggest renaming is because the latter "mark_pending," is because, as stated, a frame deemed lost does not necessarily mean that it would be remarked as pending.

kazuho:
Yes, we occasionally see failures in the end-to-end loss test. Please ignore them.

Regarding the assertion failure, maybe https://github.com/h2o/quicly/pull/306#discussion_r423314261 is the answer?

lib/quicly.c
```diff
@@ -3745,6 +3905,34 @@ static int do_send(quicly_conn_t *conn, quicly_send_context_t *s)
...
+                    for (i = 0; i < PTLS_ELEMENTSOF(conn->egress.retire_cid.sequences); i++) {
+                        uint64_t sequence = conn->egress.retire_cid.sequences[i];
+                        if (sequence == UINT64_MAX)
+                            break; /* we've sent all pending sequence numbers */
+                        if ((ret = send_retire_connection_id(conn, s, sequence)) != 0) {
+                            quicly_retire_cid_pop(&conn->egress.retire_cid, i);
+                            goto Exit;
+                        }
+                    }
+                    quicly_retire_cid_pop(&conn->egress.retire_cid, i);
```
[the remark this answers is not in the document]
hfujita: Please note that `pop` removes *multiple* items at the front. And `pop` call is located *outside* of the for loop, after knowing how many RCIDs we actually sent. This is consistent with the issued CID (`quicly_issued_cid_mark_inflight`) design which pops multiple items at once.

lib/quicly.c
```diff
@@ -3745,6 +3905,34 @@ static int do_send(quicly_conn_t *conn, quicly_send_context_t *s)
...
+                        }
+                    }
+                    quicly_issued_cid_mark_inflight(&conn->issued_cid, i);
+                    /* send RETIRE_CONNECTION_ID */
+                    for (i = 0; i < PTLS_ELEMENTSOF(conn->egress.retire_cid.sequences); i++) {
+                        uint64_t sequence = conn->egress.retire_cid.sequences[i];
+                        if (sequence == UINT64_MAX)
+                            break; /* we've sent all pending sequence numbers */
+                        if ((ret = send_retire_connection_id(conn, s, sequence)) != 0) {
+                            quicly_retire_cid_pop(&conn->egress.retire_cid, i);
```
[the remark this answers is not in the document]
hfujita: This is to remove items successfully sent in prior iterations of the loop. See https://github.com/h2o/quicly/pull/306#discussion_r423348004 on the `pop` function handling multiple items at once.

lib/quicly.c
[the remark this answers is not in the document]
kazuho: Ah I see, thank you for the clarification.

Can we change this logic to something like below? I think that would avoid the confusion that I had.

```c
for (...) {
    if ((ret = send_retire_connection_id(...)) != 0)
        break;
}
quicly_retire_cid_pop(&conn->egress.retire_cid, i);
if (ret != 0)
    goto Exit;
```

lib/retire_cid.c
```diff
@@ -0,0 +1,65 @@
...
+}
+
+void quicly_retire_cid_pop(quicly_retire_cid_set_t *set, size_t num_pop)
+{
+    assert(num_pop <= PTLS_ELEMENTSOF(set->sequences));
+    if (num_pop > set->_num_pending)
+        num_pop = set->_num_pending;
```
kazuho, line 58: I think we can change this to `assert(num_pop < set->_num_pending)`?

lib/retire_cid.c
```diff
@@ -0,0 +1,65 @@
...
+    /* move the remaining pending sequence numbers to the front */
+    memmove(set->sequences, set->sequences + num_pop, sizeof(uint64_t) * (set->_num_pending - num_pop));
+    /* insert sentinel at the end */
+    if (num_pop > 0)
```
kazuho, line 62: How about doing `if (num_pop == 0) return;` at the beginning of the function (after assertions)?

lib/retire_cid.c
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    assert(set->sequences[set->_num_pending] == UINT64_MAX);
+    set->sequences[set->_num_pending] = sequence;
+    set->_num_pending++;
+    if (set->_num_pending < PTLS_ELEMENTSOF(set->sequences))
+        set->sequences[set->_num_pending] = UINT64_MAX;
+}
+
+void quicly_retire_cid_pop(quicly_retire_cid_set_t *set, size_t num_pop)
```
kazuho, line 54: I think we might want to rename "pop" to "shift."

This is a FIFO (instead of LIFO, which push-pop implies). "shift" is the term we use for actions that removes the first entries of a vector, see `on_send_shift`.

I think using "push" for `quicly_retire_cid_push` is fine, as we are adding an entry to the end of the vector.

lib/retire_cid.c
hfujita: It is necessary because current `memmove` invocation only moves valid sequence number entries (not including the sentinel).

That being said, this part will go away because I'll be removing the sentinel as you [pointed out](https://github.com/h2o/quicly/pull/306#discussion_r423316327).

lib/retire_cid.c
hfujita: 👍 Maybe I was too contaminated with C++/STL-ism (though it might be called in `pop_front` in that case).

include/quicly/issued_cid.h
hfujita: I agree that this can be of type `void`.

Just for future reference, my understanding of the convention so far is that:
* returns boolean => 1 for success, 0 for failure
* returns error code => 0 for success, non-zero (mostly positive?) for failure

include/quicly/issued_cid.h
```diff
@@ -0,0 +1,155 @@
...
+    quicly_cid_t cid;
+    uint8_t stateless_reset_token[QUICLY_STATELESS_RESET_TOKEN_LEN];
+};
+
+/**
+ * manages a list of connection IDs we issue to the peer
+ */
+struct st_quicly_issued_cid_set_t {
+    /**
+     * storage to retain issued CIDs
```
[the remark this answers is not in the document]
hfujita: 👍 Actually, after another thought, I've renamed `_capacity` to `_size`. `_capacity` in C++ vector means the physical limit of the array, so it would have been confusion to people familiar with C++.

include/quicly/issued_cid.h
hfujita: After refactoring, this function has become a boolean, with true indicating "there is something to send".

lib/retire_cid.c
hfujita: `assert(num_pop <= set->_num_pending)`, right?

lib/retire_cid.c
hfujita: By removing sentinel this is no longer necessary.

hfujita:
@kazuho I have addressed your comments. Please take a look when you get a chance, thanks!

include/quicly/issued_cid.h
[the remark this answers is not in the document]
kazuho: Makes perfect sense. Actually, we use the two terms the same way (see `quicly_sendbuf_t::vecs` for example).

lib/quicly.c
```diff
@@ -810,6 +827,52 @@ static int schedule_path_challenge(quicly_conn_t *conn, int is_response, const u
...
+    uint64_t capacity = conn->super.peer.transport_params.active_connection_id_limit;
+    if (capacity > QUICLY_LOCAL_ACTIVE_CONNECTION_ID_LIMIT)
+        capacity = QUICLY_LOCAL_ACTIVE_CONNECTION_ID_LIMIT;
+    return capacity;
+}
+
+/**
+ * initialize a new connection ID to be sent
+ */
+static int generate_new_connection_id(quicly_conn_t *conn, quicly_issued_cid_t *cid)
```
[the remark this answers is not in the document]
hfujita: Done.

janaiyengar:
On the naming issue, I agree with changing from `received_cid`. Though, `consumed_cid` suggests that it has been used and is done, where it is simply the CID that has been received from the peer. I might suggest `issued_cid` and `peer_cid`. They're not complementary names, but I don't think they should be, because they aren't complementary functions.

kazuho:
If we are to go with "peer_cid", would it make sense to change the counterpart from "issued_cid" to "host_cid"?

Two reasons:
* I think that "peer_cid" implies that it is CIDs *issued* by peer. Having "issued_cid" and "peer_cid" could be a bit confusing.
* We occasionally use terms like "server CID" and "client CID". Using "host" and "peer" would be a natural extension of that convention.

janaiyengar:
`host_cid` is ok, but I think `local_cid` is better. That said, consistency is better than technical correctness IMO, so perhaps sticking with `host_cid` because it is what we already use might be best.

kazuho:
Thank you. So I think we have two choices:
* Use "host" and "peer", based on current convention.
* Change to "local" and "remote", including the terms we already use.

I think I'm inclined to doing the latter, because it's something used by TCP, and because it is more explicit. "host" is ambiguous because we see usages like "peer host" in literature.

janaiyengar:
I think switching to local and remote makes sense to me.

hfujita:
Push a couple of renaming patches so we use "local/remote" across the library (at least internally).

The following public APIs are not touched; if we ever decide to do it, perhaps we can do it in another PR.
```
quicly_get_peer_cid
quicly_get_peer_transport_parameters
quicly_get_peer_next_stream_id
quicly_get_peername
quicly_get_host_next_stream_id
```

hfujita:
After an offline discussion I've pushed naming changes to public API as well.

kazuho:
Merged! Thank you for all the efforts.

hfujita:
Thank you a lot for your guidance!

PR #370, opened 2020-07-06
CUBIC CC implementation

### LINKED ISSUE #365: Adding supplementary congestion control algorithms
I recently added the [CUBIC CC algorithm](https://tools.ietf.org/html/rfc8312) to an older quicly version for a project. We have since verified that it works as expected and would be happy to contribute the code back to quicly. But we figured it would be a good idea to discuss the topic in an issue first before porting the changes to quicly's master and submitting a PR. To allow applications to select between CC algorithms, a few changes to the CC interface are necessary:

1) `quicly_context_t` is amended with a new struct `quicly_cc_conf_t` to pass the desired algorithm (similar to the parameters for [loss.h](https://github.com/h2o/quicly/blob/0ab1c6723ce308538a1313382e3e3475cd6b7b36/include/quicly/loss.h#L34-L69)). This struct could also be extended to configure the algorithm, e.g., the beta and C parameters in CUBIC.

2) `quicly_cc_t` gains two new members: a pointer to a struct that represents a CC algorithm (i.e., containing pointers to the implementation of the `quicly_cc_init`, `quicly_cc_on_acked`, `quicly_cc_on_lost`, and `quicly_cc_on_persistent_congestion` functions) and a union to store algorithm-specific state (e.g., K and W_max for CUBIC)

3) New `quicly_cc_*` functions replace the ones in `cc-reno.c` to forward calls to the function pointers described above. Ideally, since different algorithms require different information about the connection, these would also be passed a (const) pointer to `quicly_conn_t`. For example, CUBIC requires the smoothed RTT estimate which is not available without pointer hacks in the interface the Reno implementation uses.

4) The old functions in `cc-reno.c` are renamed (e.g., to `quicly_reno_*`) and make up the struct for the Reno CC implementation (which continues to be the default algorithm).

I think this interface allows for the greatest flexibility going forward without any changes outside of the files implementing congestion control (other than the new context entry). I'm very interested to hear your feedback and improvements for this proposal, and whether you are interested in pursuing additional/modular congestion control algorithms at all.

PS: We also looked at implementing Google's [BBR algorithm](https://github.com/google/bbr) for our project, but that requires pacing. As far as we could tell, quicly does not support that. Are there plans to do so in the future, considering that [the spec recommends it](https://tools.ietf.org/html/draft-ietf-quic-recovery-29#section-7.9)?

TheJokr:
Resolves #365. The changes to the interfaces are mostly what I proposed there, but I had to make a few adjustments while porting the code. There are comments on the diff to explain/discuss those.


CMakeLists.txt
```diff
@@ -89,6 +90,7 @@ IF (WITH_DTRACE)
 ENDIF ()
 
 ADD_LIBRARY(quicly ${QUICLY_LIBRARY_FILES})
+TARGET_LINK_LIBRARIES(quicly LINK_PUBLIC m)
```
TheJokr, line 93: I'm much more familiar with newer CMake versions where interface dependencies are better supported. Since your minimum required version is 2.8.11, I found this solution to define them without having to set a policy. Public/interface shouldn't make much of a difference for a library.

include/quicly/cc.h
```diff
@@ -33,20 +33,52 @@ extern "C" {
...
+} quicly_cc_conf_t;
+
+#define QUICLY_CC_SPEC_CONF                                                                                                        \
+    {                                                                                                                              \
+        CC_RENO_MODIFIED /* type */                                                                                                \
+    }
+
+#define QUICLY_CC_PERFORMANT_CONF                                                                                                  \
+    {                                                                                                                              \
+        CC_RENO_MODIFIED /* type */                                                                                                \
```
TheJokr, line 65: Chose Reno for the performance conf for now for backwards compatibility and since the CUBIC implementation is new. As far as I know though most modern OS use CUBIC for TCP by default, so it might be a good idea to switch over down the line.

include/quicly/cc.h
```diff
@@ -85,23 +148,63 @@ typedef struct st_quicly_cc_t {
     uint32_t num_loss_episodes;
 } quicly_cc_t;
 
+struct st_quicly_cc_impl_t {
+    void (*quicly_cc_init)(quicly_cc_t *, const quicly_cc_conf_t *, uint32_t);
+    void (*quicly_cc_on_acked)(quicly_cc_t *, const quicly_loss_t *, uint32_t, uint64_t, uint32_t, uint32_t);
```
TheJokr, line 153: I would have preferred to use `const quicly_conn_t *` directly to make the interface more generic, but the struct is only defined in quicly.c. It would also allow the removal of the `max_udp_payload_size` (last uint32) parameter.

I understand that it shouldn't be exposed in a public header, but maybe defining it in a header next to the source files (for inclusion via `#include "..."`) would be a solution to this?

include/quicly/cc.h
```diff
@@ -85,23 +148,63 @@ typedef struct st_quicly_cc_t {
...
+        break;
+    }
+    cc->impl->quicly_cc_init(cc, conf, initcwnd);
+}
+
 /**
  * Called when a packet is newly acknowledged.
  */
-void quicly_cc_on_acked(quicly_cc_t *cc, uint32_t bytes, uint64_t largest_acked, uint32_t inflight, uint32_t max_udp_payload_size);
+static inline void quicly_cc_on_acked(quicly_cc_t *cc, const quicly_loss_t *loss, uint32_t bytes, uint64_t largest_acked,
```
TheJokr, line 184: Inlined all of these since they only forward calls anyway.

lib/cc-cubic.c
```diff
@@ -0,0 +1,155 @@
...
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+ * IN THE SOFTWARE.
+ */
+
+#include <math.h>
+#include "quicly/cc.h"
+
+#define QUICLY_MIN_CWND 2
+
+typedef double cubic_float_t;
```
TheJokr, line 29: One of the things still on my todo list is comparing this implementation to other QUIC stacks with CUBIC (mvfst, msquic) and maybe TCP stacks to see how they implement the arithmetic operations.

Storing/calculating everything as doubles probably isn't the fastest way, but the one implementation I investigated prior to starting work on this ([quiche](https://github.com/cloudflare/quiche/blob/master/src/recovery/cubic.rs)) also used 64bit floats. Open for suggestions!

lib/cc-cubic.c
```diff
@@ -0,0 +1,155 @@
...
+
+    /* Congestion avoidance. */
+    cubic_float_t t_sec = calc_cubic_t(cc);
+    cubic_float_t rtt_sec = loss->rtt.smoothed / (cubic_float_t)1000; /* ms -> s */
+
+    uint32_t w_cubic = calc_w_cubic(cc, t_sec, max_udp_payload_size);
+    uint32_t w_est = calc_w_est(cc, t_sec, rtt_sec, max_udp_payload_size);
+
+    if (w_cubic < w_est) {
+        /* 4.2 TCP-Friendly Region */
```
TheJokr, line 93: Equation and section numbers refer to the [CUBIC RFC](https://tools.ietf.org/html/rfc8312)

include/quicly/cc.h
kazuho: I think the other question here is if we want to have unit tests for CC. If the answer is yes, then not having dependency from CC to `quicly_conn_t` is highly beneficial.

I think that having no unit test for Reno / Cubic is fine, but the importance of unit tests rise as the we add complex CC. For that, I might prefer retaining isolation.

lib/cc-cubic.c
kazuho: I'm fine with starting with a double. I'd assume that most platforms capable of running crypto at certain speed would not suffer from us doing FP in CC.

lib/cc-cubic.c
```diff
@@ -0,0 +1,155 @@
...
+
+static void quicly_cubic_init(quicly_cc_t *cc, const quicly_cc_conf_t *conf, uint32_t initcwnd)
+{
+    cc->cwnd = cc->cwnd_initial = cc->cwnd_maximum = initcwnd;
+    cc->ssthresh = cc->cwnd_minimum = UINT32_MAX;
+}
+
+/* TODO: Avoid increase if sender was application limited. */
+static void quicly_cubic_on_acked(quicly_cc_t *cc, const quicly_loss_t *loss, uint32_t bytes, uint64_t largest_acked,
+                                  uint32_t inflight, uint32_t max_udp_payload_size)
```
kazuho, line 70: We can remove `quicly_` from static functions being defined, as they do not pollute the global namespace. We can remove `cubic_` too, though I am fine with retaining it as it might turn out to be helpful when we look at stack traces.

include/quicly/cc.h
kazuho: For simplicity, maybe call cc->impl->xxx directly from lib/quicly.c?

include/quicly/cc.h
TheJokr: Good point, I hadn't actually considered unit tests. 

Adding more parameters to the functions in the future is pretty simple, and certainly not everything in `quicly_conn_t` is relevant for CC. That raises the question of what level of abstraction to choose though, e.g., CUBIC only uses the RTT estimate from `quicly_loss_t`. I chose to include the entire struct since other CCs might use different parts.

include/quicly/cc.h
TheJokr: Agreed - this was a leftover from an old version to avoid changes outside the cc source files. But once I had to add new parameters, that became necessary anyway.

include/quicly/cc.h
```diff
@@ -33,20 +33,52 @@ extern "C" {
...
+     */
+    CC_CUBIC
 } quicly_cc_type_t;
 
+typedef struct st_quicly_cc_conf_t {
+    /**
+     * Congestion controller type.
+     */
+    quicly_cc_type_t type;
+} quicly_cc_conf_t;
```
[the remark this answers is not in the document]
TheJokr: Currently, the `quicly_cc_conf_t` struct indeed serves little purpose. I added it to allow for more configuration options in `quicly_context_t` later, like setting Renos/CUBICs beta value.

You are suggesting to pass a `quicly_cc_impl_t` pointer in `quicly_context_t`, right? That would indeed eliminate the translation, but makes similar configuration parameters harder to implement for users. If they were included in impl, one would have to create a new struct (that has to live as long as the connection) to set those parameters.

What do you think about keeping `quicly_cc_conf_t`, but replacing `quicly_cc_type_t` with a `quicly_cc_impl_t` pointer? Then users can even supply their own CC implementation. type in `quicly_cc_t` would be moved into `quicly_cc_impl_t`. And configuration parameters could still be added later.

include/quicly/cc.h
kazuho: I agree that the level of abstraction is the concern. That would continue to be, as we change things. Eventually, we might end up in having an set of callbacks that provide the necessary information.

But for the time being I think what we have in the PR is good.

lib/cc-cubic.c
```diff
@@ -0,0 +1,155 @@
...
+#define QUICLY_MIN_CWND 2
+
+typedef double cubic_float_t;
+#define QUICLY_CUBIC_C ((cubic_float_t)0.4)
+#define QUICLY_CUBIC_BETA ((cubic_float_t)0.7)
+
+/* Calculates the elapsed time since the last congestion event (parameter t) */
+static cubic_float_t calc_cubic_t(const quicly_cc_t *cc)
+{
+    cubic_float_t clock_delta = clock() - cc->state.cubic.avoidance_start;
```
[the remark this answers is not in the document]
TheJokr: I didn't search for an existing clock functionality in quicly before implementing CUBIC. As far as I can see, loss detection is passed the current time via the now parameter, set to `conn->stash.now`. I guess this would be the appropriate time source to use? And its resolution appears to be ms?

include/quicly/cc.h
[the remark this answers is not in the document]
kazuho: I do not have a strong opinion if we need to do everything now in this PR, but I would assume that we'd eventually end up in using factory method pattern. As you correctly point out, it is beneficial to have a way to provide CC the configuration parameters. At the same time, each CC might require different set of parameters.

That natural outcome of these requirements is IMO the use of a factory method. The *public* API would look like this:

```c
typedef struct st_quicly_conn_t {
    ...
    /**
     * Creates a congestion controller for given connection.
     */
    quicly_cc_t (*init_cc)(quicly_conn_t *conn);
    ...
} quicly_conn_t;

void quicly_cc_reno_init(quicly_cc_t *cc, uint32_t initcwnd);
void quicly_cc_cubic_init(quicly_cc_t *cc, uint32_t initcwnd, ... /* add more parameters as we need */);
```

There would no longer be a `cc_init` callback in `quicly_cc_impl_t`.

Each user is expected to supply their own `init_cc` callback (we can have defaults associated to the default contexts), that calls the CC they want to use, with the parameters they want to use.

lib/cc-cubic.c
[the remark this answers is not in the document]
kazuho: Correct.

include/quicly/cc.h
TheJokr: Left `quicly_cc_init` as is for now since there is more logic in there.

lib/cc-cubic.c
TheJokr: Left `cubic_`/`reno_` in for now. Makes searching easier as well.

include/quicly/cc.h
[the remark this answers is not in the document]
TheJokr: I like that! As for doing everything now: especially when it comes to (public) configuration options, I prefer to avoid foreseeable breaking changes.

Will take a shot at implementing this tomorrow. Since `quicly_conn_t` isn't publicly exposed, I'd suggest to pass `quicly_cc_t *` (to modify) + `const quicly_context_t *` (to calculate initcwnd, ...) instead. Not sure what the return value would be used for?

lib/cc-cubic.c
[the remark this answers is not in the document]
TheJokr: Will implement tomorrow.

include/quicly/cc.h
TheJokr: Init function is also gone now due to new config interface.

include/quicly/cc.h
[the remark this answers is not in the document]
TheJokr: Implemented now. Used the existing `QUICLY_CALLBACK_TYPE` macro for init_cc to allow users to pass additional data. The defaults are now located in `default.h/.c` (like the other default callback structs).

lib/cc-cubic.c
TheJokr: It appears that msquic is the only implementation out of the ones I mentioned that uses integer arithmetic for CUBIC. mvfst has TODOs for benchmarking different cbrt implementations against each other, so maybe their decision will provide some insight. After putting some more thought into this I don't think the effort is currently worth it.

TheJokr:
Rebased to current master.

polymagicien:
Even though that seems to work great, why not choose to implement the pseudo code of the "CUBIC: A New TCP-Friendly High-Speed TCP Variant" paper, written by the authors of the RFC8312 ? Just for information, it is the way CUBIC is implemented in QUIC-GO and Quiche (chromium).

lib/cc-cubic.c
```diff
@@ -0,0 +1,156 @@
...
+
+    /* Congestion avoidance. */
+    cubic_float_t t_sec = calc_cubic_t(cc, now);
+    cubic_float_t rtt_sec = loss->rtt.smoothed / (cubic_float_t)1000; /* ms -> s */
+
+    uint32_t w_cubic = calc_w_cubic(cc, t_sec, max_udp_payload_size);
+    uint32_t w_est = calc_w_est(cc, t_sec, rtt_sec, max_udp_payload_size);
+
+    if (w_cubic < w_est) {
+        /* 4.2 TCP-Friendly Region */
```
janaiyengar, line 87:
```suggestion
        /* RFC 8312, Section 4.2; TCP-Friendly Region */
```

lib/cc-cubic.c
```diff
@@ -0,0 +1,156 @@
...
+        /* Prevent cwnd from shrinking if W_est is reduced due to RTT increase */
+        if (w_est > cc->cwnd) {
+            cc->cwnd = w_est;
+            if (cc->cwnd_maximum < cc->cwnd)
+                cc->cwnd_maximum = cc->cwnd;
```
janaiyengar, line 92: You can move these two lines to the end of the function, and remove it from the else clause as well below.

lib/cc-cubic.c
```diff
@@ -0,0 +1,156 @@
...
+        }
+    } else {
+        /* 4.3/4.4 CUBIC Region */
```
janaiyengar, line 95:
```suggestion
        /* RFC 8312, Section 4.3/4.4; CUBIC Region */
```

lib/cc-cubic.c
```diff
@@ -0,0 +1,156 @@
...
+        /* Prevent cwnd from shrinking after fast convergence (beta*W_max < cwnd) */
+        cubic_float_t increment = calc_w_cubic(cc, t_sec + rtt_sec, max_udp_payload_size);
```
janaiyengar, line 97:
```suggestion
        cubic_float_t w_cubic_target = calc_w_cubic(cc, t_sec + rtt_sec, max_udp_payload_size);
```

lib/cc-cubic.c
```diff
@@ -0,0 +1,156 @@
...
+        if (increment > cc->cwnd) {
```
janaiyengar, line 98:
```suggestion
        if (w_cubic_target > cc->cwnd) {
```

lib/cc-cubic.c
```diff
@@ -0,0 +1,156 @@
...
+            /* (W_cubic(t+RTT) - cwnd)/cwnd * MSS = (W_cubic(t+RTT)/cwnd - 1) * MSS */
+            increment = ((increment / cc->cwnd) - 1) * max_udp_payload_size;
+            cc->cwnd += increment;
```
janaiyengar, line 101:
```suggestion
            cc->cwnd += ((w_cubic_target / cc->cwnd) - 1) * max_udp_payload_size;
```

lib/cc-cubic.c
```diff
@@ -0,0 +1,156 @@
...
+        /* 4.2 TCP-Friendly Region */
+        /* Prevent cwnd from shrinking if W_est is reduced due to RTT increase */
+        if (w_est > cc->cwnd) {
+            cc->cwnd = w_est;
+            if (cc->cwnd_maximum < cc->cwnd)
+                cc->cwnd_maximum = cc->cwnd;
+        }
+    } else {
+        /* 4.3/4.4 CUBIC Region */
+        /* Prevent cwnd from shrinking after fast convergence (beta*W_max < cwnd) */
```
janaiyengar, line 96: How is this prevented in the code here?

lib/cc-cubic.c
```diff
@@ -0,0 +1,156 @@
...
+    cc->recovery_end = next_pn;
+
+    ++cc->num_loss_episodes;
+    if (cc->cwnd_exiting_slow_start == 0)
+        cc->cwnd_exiting_slow_start = cc->cwnd;
+
+    cc->state.cubic.avoidance_start = now;
+    cc->state.cubic.w_max = cc->cwnd;
+
+    /* 4.6 Fast Convergence */
```
janaiyengar, line 123:
```suggestion
    /* RFC 8312, Section 4.6; Fast Convergence */
```

lib/cc-cubic.c
```diff
@@ -0,0 +1,156 @@
...
+    /* Zero initialisation ensures this condition is false during slow start */
+    if (cc->state.cubic.w_max < cc->state.cubic.w_last_max) {
+        cc->state.cubic.w_last_max = cc->state.cubic.w_max;
+        cc->state.cubic.w_max *= (1.0 + QUICLY_CUBIC_BETA) / 2.0;
+    } else {
+        cc->state.cubic.w_last_max = cc->state.cubic.w_max;
+    }
+    update_cubic_k(cc, max_udp_payload_size);
+
+    /* 4.5 Multiplicative Decrease */
```
janaiyengar, line 133:
```suggestion
    /* RFC 8312, Section 4.5; Multiplicative Decrease */
```

lib/cc-cubic.c
```diff
@@ -0,0 +1,156 @@
...
+
+    ++cc->num_loss_episodes;
+    if (cc->cwnd_exiting_slow_start == 0)
+        cc->cwnd_exiting_slow_start = cc->cwnd;
+
+    cc->state.cubic.avoidance_start = now;
+    cc->state.cubic.w_max = cc->cwnd;
+
+    /* 4.6 Fast Convergence */
+    /* Zero initialisation ensures this condition is false during slow start */
```
janaiyengar, line 124:
```suggestion
    /* w_last_max is initialized to zero; therefore this condition is false when exiting slow start */
```

lib/cc-cubic.c
```diff
@@ -0,0 +1,156 @@
...
+#define QUICLY_MIN_CWND 2
+
+typedef double cubic_float_t;
+#define QUICLY_CUBIC_C ((cubic_float_t)0.4)
+#define QUICLY_CUBIC_BETA ((cubic_float_t)0.7)
+
+/* Calculates the time elapsed since the last congestion event (parameter t) */
+static cubic_float_t calc_cubic_t(const quicly_cc_t *cc, int64_t now)
+{
+    cubic_float_t clock_delta = now - cc->state.cubic.avoidance_start;
```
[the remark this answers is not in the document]
janaiyengar: @kazuho pointed out to me offline that there is an invariant in `now`: `quicly_context_t::now_cb` is guaranteed to return an increasing value across calls. I think that covers it and this assert is not required.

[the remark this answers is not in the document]
kazuho: :+1: Just FWIW, I have make it clear that we have such design contract in bdbcbfd.

lib/cc-cubic.c
```diff
@@ -0,0 +1,156 @@
...
+    cubic_float_t tk = t_sec - cc->state.cubic.k;
+    return (QUICLY_CUBIC_C * (tk * tk * tk) * max_udp_payload_size) + cc->state.cubic.w_max;
+}
+
+/* Equation 2 */
+/* K depends solely on W_max, so we update both together on congestion events */
+static void update_cubic_k(quicly_cc_t *cc, uint32_t max_udp_payload_size)
+{
+    cubic_float_t w_max_mss = cc->state.cubic.w_max / (cubic_float_t)max_udp_payload_size;
+    cc->state.cubic.k = cbrt(w_max_mss * ((1 - QUICLY_CUBIC_BETA) / QUICLY_CUBIC_C));
```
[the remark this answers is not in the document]
TheJokr: I put them there to allow for constant folding. Left-to-right associativity + floating point error would prevent this otherwise, I think.

lib/cc-cubic.c
TheJokr: Through the check for `increment > cc->cwnd`. Otherwise, for t close to 0 after fast convergence, `W_cubic(t) = beta * W_max = beta * 0.85 * W_last_max` < `beta * W_last_max = cwnd`.

Similarly, W_est(t) can shrink when the estimated RTT changes, hence there is a similar check there.

TheJokr:
Thanks to both of you for your reviews!

@polymagicien I prefer to use RFCs over academic papers as references for implementations, as the former generally represent the consensus of many researchers and implementers on the documented topic. E.g., in this case, the RFC was published 10 years after the paper (more operational experience) and describes the algorithm in more general terms than the Linux kernel-specific pseudo code in the paper. Additionally, the RFC is more easily accessible to anyone and much more concise.

lib/cc-cubic.c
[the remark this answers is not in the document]
janaiyengar: Ah yes, that makes sense. Let's leave it as it is.

janaiyengar:
@TheJokr : Have you thought about tests for this? I know we don't have any for Reno as well, but if you've verified the performance of this implementation, do you have any tests that could be built from your verification?

polymagicien:
@TheJokr : I submitted a pull request on `TheJokr:cubic-cc` adding a parameter called `num_connections_`. It as an impact on the way the cwnd shrinks / increases, and aims at making the competition with other flows fairer. It is particularly useful when the QUIC connection aggregates TCP connections in different streams, as mentionned in the "CUBIC: A New TCP-Friendly High-Speed TCP Variant" paper.

TheJokr:
@janaiyengar I verified it experimentally using netns + qdiscs and plotting the cwnd curve afterwards, so my process is not automated at all. If I were to write unit tests for CUBIC, they would more or less amount to verifying that the algorithm is in the correct region and that the calculations produce the correct results. This would probably also mean lots of overlap with Reno tests (slow start, loss reduction, ...), so maybe it's a better idea to tackle this in a separate PR for Reno and CUBIC combined.

@polymagicien I appreciate your input + PR, but I feel like discussing this enhancement in a separate PR on this repo (once CUBIC is merged) would be a better idea: whether to make it build-/run-time configurable (or automatically detected), what default/constant to use, etc. This decision should lie with quicly's maintainers, therefore I closed your PR on my fork.

lib/cc-cubic.c
TheJokr: Extended the comment now.

janaiyengar:
On the multi-connection idea -- I agree with @kazuho. It's not a direction I think we should go in.

PR #378, opened 2020-07-31
rename a USDT param name to follow `void* <name>` and `size_t <name>_len` convention

gfx:
This is because `h2olog` uses heuristics that `void* <name>` must be paired with `size_t <name>_len`.


gfx:
Thanks. BTW, I have no permission to merge PRs. Can I have it?

PR #402, opened 2020-09-26
actually use the certs provided by the interop runner

marten-seemann:


janaiyengar (approved):
Ugh, sorry I missed this too.

PR #403, opened 2020-09-29
Preliminary work to produce QLog compliant traces

toru:
Hello. This PR aims to bootstrap a mechanism to compute [QLog](https://github.com/quiclog/internet-drafts) compliant QUIC server traces. Starting small by focusing on just the `packet_received` and `packet_sent` events.

### DIFF
--- a/lib/quicly.c
+++ b/lib/quicly.c
@@ -2933,10 +2933,8 @@ static int commit_send_packet(quicly_conn_t *conn, quicly_send_context_t *s, enu
         quicly_sentmap_commit(&conn->egress.loss.sentmap, (uint16_t)packet_bytes_in_flight);
 
     conn->egress.cc.impl->cc_on_sent(&conn->egress.cc, &conn->egress.loss, (uint32_t)packet_bytes_in_flight, conn->stash.now);
-    QUICLY_PROBE(PACKET_COMMIT, conn, conn->stash.now, conn->egress.packet_number, s->dst - s->target.first_byte_at,
-                 !s->target.ack_eliciting);
-    QUICLY_PROBE(QUICTRACE_SENT, conn, conn->stash.now, conn->egress.packet_number, s->dst - s->target.first_byte_at,
-                 get_epoch(*s->target.first_byte_at));
+    QUICLY_PROBE(PACKET_SENT, conn, conn->stash.now, conn->egress.packet_number, s->dst - s->target.first_byte_at,
+                 get_epoch(*s->target.first_byte_at), !s->target.ack_eliciting);
 
     ++conn->egress.packet_number;
     ++conn->super.stats.num_packets.sent;
--- a/lib/quicly.c
+++ b/lib/quicly.c
@@ -5392,8 +5392,7 @@ int quicly_accept(quicly_conn_t **conn, quicly_context_t *ctx, struct sockaddr *
 
     QUICLY_PROBE(ACCEPT, *conn, (*conn)->stash.now,
                  QUICLY_PROBE_HEXDUMP(packet->cid.dest.encrypted.base, packet->cid.dest.encrypted.len), address_token);
-    QUICLY_PROBE(CRYPTO_DECRYPT, *conn, (*conn)->stash.now, pn, payload.base, payload.len);
-    QUICLY_PROBE(QUICTRACE_RECV, *conn, (*conn)->stash.now, pn);
+    QUICLY_PROBE(PACKET_RECEIVED, *conn, (*conn)->stash.now, pn, payload.base, payload.len);
 
     /* handle the input; we ignore is_ack_only, we consult if there's any output from TLS in response to CH anyways */
     (*conn)->super.stats.num_packets.received += 1;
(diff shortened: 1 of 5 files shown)


lib/quicly.c
```diff
@@ -5586,12 +5583,11 @@ int quicly_receive(quicly_conn_t *conn, struct sockaddr *dest_addr, struct socka
     if ((ret = decrypt_packet(header_protection, aead.cb, aead.ctx, &(*space)->next_expected_packet_number, packet, &pn,
                               &payload)) != 0) {
         ++conn->super.stats.num_packets.decryption_failed;
-        QUICLY_PROBE(CRYPTO_DECRYPT, conn, conn->stash.now, pn, NULL, 0);
+        QUICLY_PROBE(CRYPTO_DECRYPT_FAILURE, conn, conn->stash.now, pn);
```
kazuho, line 5586: Looking at https://quiclog.github.io/internet-drafts/draft-marx-qlog-event-definitions-quic-h3.html#name-packet_dropped, I think we might want to make this something like below (assuming that the document I'm referring to is the right one), but we can do that later if that's your preference.

```suggestion
        QUICLY_PROBE(PACKET_DROPPED, conn, conn->stash.now, "payload_decrypt_error", pn);
```

kazuho (commented):
Thank you for working on this PR. The proposed changes look good to me, assuming this it is part of a WIP.

lib/quicly.c
toru: No, this is an excellent suggestion. Thank you, and I'll take it. In fact it's a bit embarrassing that I did not look at the qlog event spec for inspiration on the new name.

toru:
> The proposed changes look good to me, assuming this it is part of a WIP.

@kazuho Yup, it's a WIP of a much larger problem. At first I was thinking about submitting a series of PRs, but I'm starting to think otherwise. Would it make more sense if I continue committing to this branch/PR and ask you to review as it grows?

janaiyengar (commented):
+1 to @kazuho's suggestion, but this looks great. Thanks for working on it @toru!
(and apologies for taking my time to get to the review)

kazuho:
> Would it make more sense if I continue committing to this branch/PR and ask you to review as it grows?

I think we might actually prefer that. From the perspective of using the traces, it might be better to land things at once (at least the changes that constitute MVP).

toru:
@kazuho @janaiyengar I think now's a good time to ask for an interim review. The big difference since your last review is the introduction of this script: [qlog-adapter.py](https://github.com/h2o/quicly/blob/toru/packet-received-probe/misc/qlog-adapter.py) -- The design of the program is similar to Quicly's frame handler architecture. Each QLog event is given a handler function:

```python
QLOG_EVENT_HANDLERS = {
    "packet-received": handle_packet_received,
    "packet-sent": handle_packet_sent
}
```

Similarly, Quicly events that carry frame information are also given a dedicated handler function. These functions extract the frame information out of an event structure, and organizes it in a QLog compliant way.

```python
FRAME_EVENT_HANDLERS = {
    "ack-send": handle_ack_send,
    "ping-receive": handle_ping_receive,
    "quictrace-recv-ack": handle_quictrace_recv_ack,
    "recv-crypto-frame": handle_recv_crypto_frame,
}
```

Given these two Handler Registries, the main function only needs to be concerned with:

- CLI options handling
- Read the h2olog QUIC trace file
- Loop over the QUIC trace events and lookup the registry, such that:

```python
for i, event in enumerate(source_events):
    handler = QLOG_EVENT_HANDLERS.get(event["type"])
    if handler:
        trace["events"].append(handler(source_events, i))
```

The output so far is well received by the QVis toolsuite. Here's a cropped rendering from the sequence diagram tool.

qvis-sequence

## Next Step

- Implement more frame handlers
- Rename probes relevant to this iteration of the project

misc/qlog-adapter.py
```diff
@@ -0,0 +1,142 @@
+#!/usr/bin/env python
```
gfx, line 1: Could you use python3?

kazuho:
@toru Nice work! I'm excited to see us being able to use qvis.

The PR looks good to me, I'm looking forward to seeing you make further progress.

misc/qlog-adapter.py
toru: Done: d3280a8

toru:
@kazuho @janaiyengar I think this PR is ready to wrap up. I walked an extra mile and supported all the frames that we can from existing probes. Pleasantly, it turns out most of them were supportable. Here's what the QVis sequence diagram looks like with the latest code, using the same data from an earlier rendering. Much more comprehensive :)

torumk com qvis

lib/quicly.c
```diff
@@ -5586,12 +5583,11 @@ int quicly_receive(quicly_conn_t *conn, struct sockaddr *dest_addr, struct socka
     if ((ret = decrypt_packet(header_protection, aead.cb, aead.ctx, &(*space)->next_expected_packet_number, packet, &pn,
                               &payload)) != 0) {
         ++conn->super.stats.num_packets.decryption_failed;
-        QUICLY_PROBE(CRYPTO_DECRYPT, conn, conn->stash.now, pn, NULL, 0);
+        QUICLY_PROBE(PACKET_DROPPED, conn, conn->stash.now, "payload_decrypt_error", pn);
```
kazuho, line 5586: I think we might want to have a dedicated probe for decryption failure rather than using a string that identifies a cause. This is an important event that we also count, it's nice to have a way of just probing those.

Something like: `QUICLY_PROBE(PACKET_DECRYPTION_FAILURE, conn, conn->stash.now, pn)`.

kazuho (commented):
Thank you for all the efforts. Looks pretty good, just one point below.

lib/quicly.c
toru: Sounds good to me. Addressed in: 3d30865

kazuho (approved):
Thank you for the changes. Please merge once the CI succeeds.

PR #422, opened 2020-12-02
Fix an ACK frame interpretation bug in the QLog adapter

toru:
The qlog-adapter program incorrectly renders each incoming `quictrace_recv_ack` event as an individual ACK frame. The fix is to buffer acknowledged blocks until the entire ACK frame is processed. I also renamed `quictrace_recv_ack` to `ack_block_received` in order to make the purpose of the probe more clear.

Screen Shot 2020-12-01 at 5 54 50 PM

### DIFF
--- a/lib/quicly.c
+++ b/lib/quicly.c
@@ -4691,7 +4691,7 @@ static int handle_ack_frame(quicly_conn_t *conn, struct st_quicly_handle_payload
         /* Ack blocks are organized in the ACK frame and consequently in the ack_block_lengths array from the largest acked down.
          * Processing acks in packet number order requires processing the ack blocks in reverse order. */
         uint64_t pn_block_max = pn_acked + frame.ack_block_lengths[gap_index] - 1;
-        QUICLY_PROBE(ACK_BLOCK_RECV, conn, conn->stash.now, pn_acked, pn_block_max);
+        QUICLY_PROBE(ACK_BLOCK_RECEIVED, conn, conn->stash.now, pn_acked, pn_block_max);
         while (quicly_sentmap_get(&iter)->packet_number < pn_acked)
             quicly_sentmap_skip(&iter);
         do {
--- a/lib/quicly.c
+++ b/lib/quicly.c
@@ -4691,7 +4691,7 @@ static int handle_ack_frame(quicly_conn_t *conn, struct st_quicly_handle_payload
         /* Ack blocks are organized in the ACK frame and consequently in the ack_block_lengths array from the largest acked down.
          * Processing acks in packet number order requires processing the ack blocks in reverse order. */
         uint64_t pn_block_max = pn_acked + frame.ack_block_lengths[gap_index] - 1;
-        QUICLY_PROBE(QUICTRACE_RECV_ACK, conn, conn->stash.now, pn_acked, pn_block_max);
+        QUICLY_PROBE(ACK_BLOCK_RECV, conn, conn->stash.now, pn_acked, pn_block_max);
         while (quicly_sentmap_get(&iter)->packet_number < pn_acked)
             quicly_sentmap_skip(&iter);
         do {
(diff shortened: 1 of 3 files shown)


lib/quicly.c
```diff
@@ -4691,7 +4691,7 @@ static int handle_ack_frame(quicly_conn_t *conn, struct st_quicly_handle_payload
         /* Ack blocks are organized in the ACK frame and consequently in the ack_block_lengths array from the largest acked down.
          * Processing acks in packet number order requires processing the ack blocks in reverse order. */
         uint64_t pn_block_max = pn_acked + frame.ack_block_lengths[gap_index] - 1;
-        QUICLY_PROBE(QUICTRACE_RECV_ACK, conn, conn->stash.now, pn_acked, pn_block_max);
+        QUICLY_PROBE(ACK_BLOCK_RECEIVED, conn, conn->stash.now, pn_acked, pn_block_max);
```
janaiyengar, line 4694:
```suggestion
        QUICLY_PROBE(ACK_BLOCK_RECEIVE, conn, conn->stash.now, pn_acked, pn_block_max);
```

just to be consistent with other RECEIVE events

misc/qlog-adapter.py
```diff
@@ -27,10 +27,21 @@
...
     frames = []
+    acked = []
     for i in range(idx+1, len(events)):
         ev = events[i]
         if ev["type"] == "packet-prepare" or ev["type"] in QLOG_EVENT_HANDLERS:
             break
+
+        # An ACK frame can't be re-composed in an iteration. Continue buffering
+        # the ACK blocks until all the blocks are processed.
+        if ev["type"] == "ack-block-received":
```
janaiyengar, line 38:
```suggestion
        if ev["type"] == "ack-block-receive":
```

misc/qlog-adapter.py
```diff
@@ -27,10 +27,21 @@
...
+            acked.append([ev["ack-block-begin"],ev["ack-block-end"]])
+            continue
+        elif ev["type"] == "quictrace-recv-ack-delay":
```
janaiyengar, line 41:
```suggestion
        elif ev["type"] == "ack-delay-receive":
```

Rename this PROBE as well

quicly-probes.d
```diff
@@ -54,6 +54,7 @@ provider quicly {
                           size_t inflight);
     probe cc_congestion(struct st_quicly_conn_t *conn, int64_t at, uint64_t max_lost_pn, size_t inflight, uint32_t cwnd);
 
+    probe ack_block_received(struct st_quicly_conn_t *conn, int64_t at, uint64_t ack_block_begin, uint64_t ack_block_end);
```
janaiyengar, line 57:
```suggestion
    probe ack_block_receive(struct st_quicly_conn_t *conn, int64_t at, uint64_t ack_block_begin, uint64_t ack_block_end);
```

janaiyengar (approved):
A couple of nits, but LGTM - thanks for finding and fixing this bug!

lib/quicly.c
toru: So the use of past tense here was intentional. A week or so ago @hfujita raised a question about using past tense, which @kazuho and I agreed because the use of past tense would follow QLog's naming convention. 84d322d is an example result of this.

misc/qlog-adapter.py
toru: Good point! will do.

kazuho (approved):
LGTM modulo naming issues, to which I'm in favor of using past tense following the convention of qlog.

misc/qlog-adapter.py
toru: Done: d8ed2fa

quicly-probes.d
toru: Resolving in favor of past-tense as mentioned in the other comment.

toru:
Thank you for the review folks. Will merge once CI passes.

lib/quicly.c
janaiyengar: I agree that it is better. We should make this consistent though, since we now have _SEND, _SENT, _RECEIVE as various probe names.

PR #424, opened 2020-12-08
Implement the "packet_lost" QLog event

toru:
Hello. This PR adds QLog's [packet_lost recovery event](https://tools.ietf.org/id/draft-marx-qlog-event-definitions-quic-h3-02.html#name-packet_lost) support to our QLog adapter program. In the process I made couple of tweaks to the BPF probes.

1. Removed `QUICTRACE_LOST` which was identical to `PACKET_LOST`
2. Made the `PACKET_LOST` probe emit the epoch byte as `packet_type`

Rendering the actual `packet_lost` event was trivial thanks to the handler registry architecture.


toru:
This PR is missing an interesting piece of information, which is the lost frames that were carried by the lost packet. After talking to @janaiyengar, I think I can reconstruct it.

kazuho (approved):

PR #430, opened 2021-01-26
Prevent potential NULL dereference

hfujita:
`hp_ctx` is not guaranteed to be non-NULL.

Fixes Coverity CID 1501098


kazuho (approved):
Nice catch! The only case we hit this problem is when allocation of a small chunk of memory returns NULL (which does not happen on linux), but we need this.

PR #454, opened 2021-05-24
Run git submodule update by cmake

uk-ar:
To simplify the workflow and reduce human errors, I propose to run git submodule update by CMake.


uk-ar:
I have understood the policy through discussion https://github.com/h2o/h2o/pull/2711#issuecomment-850029460 and will withdraw this PR. Thank you very much.

PR #459, opened 2021-06-02
client app: added common udp_wait_readready function

maxsharabayko:
There is a duplicated code in the client application, which implements waiting for read-readiness on a UDP socket.
This code is combined into a single function to reduce code duplication and cognitive complexity.

```c++
udp_wait_readready(int fd, quicly_conn_t** connections, size_t num_connections, int64_t client_enqueue_rq_at)
```

### DIFF
--- a/src/cli.c
+++ b/src/cli.c
@@ -499,6 +499,51 @@ static void on_receive_datagram_frame(quicly_receive_datagram_frame_t *self, qui
         quicly_send_datagram_frames(conn, &payload, 1);
 }
 
+/// @brief Waut on UDP socket fd for read-readiness.
+/// @param fd UDP socket descriptor.
+/// @param connections array of connections to handle. The name must not conflict with global 'conns' var.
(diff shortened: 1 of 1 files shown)


src/cli.c
```diff
@@ -499,6 +499,51 @@ static void on_receive_datagram_frame(quicly_receive_datagram_frame_t *self, qui
...
+            if (conn_to < timeout_at)
+                timeout_at = conn_to;
+        }
+
+        // client enqueues a new request to send.
+        if (client_enqueue_rq_at < timeout_at)
+            timeout_at = client_enqueue_rq_at;
+
+        if (timeout_at != INT64_MAX) {
+            int64_t delta = timeout_at - ctx.now->cb(ctx.now);
```
maxsharabayko, line 526: Here in case of client the connection context was used instead of the global `ctx`. Does it make any difference?
```c
quicly_context_t *ctx = quicly_get_context(conn);
```

PR #462, opened 2021-06-14
Preserve sentmap integrity when the `acked` callback returns an error

deweerdt:
Before this PR, we could exit `quicly_sentmap_update` with the packet
entry removed, but with frames left.
This change makes sure that the sentmap data remains usable after an
error


deweerdt:
We should fix callers to quicly_sentmap_update instead

PR #508, opened 2022-03-18
qlog-adapter: accept stdin

hfujita:
It would be natural for this type of tool to be used like
```
cmd-emits-h2olog | qlog-adapter.py > qlog-file
```
This patch modifies qlog-adapter's behavior so that if an input file name is omitted, it reads input from stdin.

### DIFF
--- a/misc/qlog-adapter.py
+++ b/misc/qlog-adapter.py
@@ -351,8 +351,7 @@ def main():
         try:
             infile = open(fn, "r")
         except OSError as e:
-            print("Failed to open %s: %s" % (fn, e.strerror))
-            sys.exit(1)
+            sys.exit("Failed to open %s: %s" % (fn, e.strerror))
 
     source_events = load_quicly_events(infile)
     print(json.dumps({
--- a/misc/qlog-adapter.py
+++ b/misc/qlog-adapter.py
@@ -328,7 +328,7 @@ FRAME_EVENT_HANDLERS = {
 def usage():
     print(r"""
 Usage:
-    python qlog-adapter.py [inTrace.jsonl]
+    python3 qlog-adapter.py [inTrace.jsonl]
 
     If the argument is omitted, inTrace will be read from stdin.
 """.strip())
--- a/misc/qlog-adapter.py
+++ b/misc/qlog-adapter.py
@@ -328,7 +328,7 @@ FRAME_EVENT_HANDLERS = {
 def usage():
     print(r"""
 Usage:
-    python3 qlog-adapter.py [inTrace.jsonl]
+    ./misc/qlog-adapter.py [inTrace.jsonl]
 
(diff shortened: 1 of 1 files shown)


misc/qlog-adapter.py
```diff
@@ -328,22 +328,32 @@ def render_ack_frame(ranges):
...
         usage()
         sys.exit(1)
 
-    (_, infile) = sys.argv
+    if len(sys.argv) == 1:
+        infile = sys.stdin
+    else:
+        (_, fn) = sys.argv
+        try:
+            infile = open(fn, "r")
```
hfujita, line 352: Just realized that I'm not closing `infile` in this case (whereas in the original code it automatically does with `with`. However I'd say I don't care because it is opened for read (no fear of data loss due to buffer not being flushed) and one time thing in a short-lived command-line utility (no fear of leaked resource being accumulated).

misc/qlog-adapter.py
```diff
@@ -328,22 +328,32 @@ def render_ack_frame(ranges):
 def usage():
     print(r"""
 Usage:
-    python qlog-adapter.py inTrace.jsonl
+    python3 qlog-adapter.py [inTrace.jsonl]
```
gfx, line 331: [nits] I'd remove "python3" from here. We should execute the script directly.

misc/qlog-adapter.py
```diff
@@ -328,22 +328,32 @@ def render_ack_frame(ranges):
...
-    (_, infile) = sys.argv
+    if len(sys.argv) == 1:
+        infile = sys.stdin
+    else:
+        (_, fn) = sys.argv
+        try:
+            infile = open(fn, "r")
+        except OSError as e:
+            print("Failed to open %s: %s" % (fn, e.strerror))
+            sys.exit(1)
```
gfx, line 355: `sys.exit("Failed to open %s: %s" % (fn, e.strerror))` seems better to write error messages to stdout and exit with non-zero status code.

gfx (approved):
It makes real sense! Thank you. LGTM except for small suggestions.

misc/qlog-adapter.py
gfx, line 331:
```suggestion
    ./misc/qlog-adapter.py [inTrace.jsonl]
```

misc/qlog-adapter.py
hfujita: Just learned about this syntax. Addressed in [eaae916](https://github.com/h2o/quicly/pull/508/commits/eaae916f290222d43e66ee5a303a8b2244471923)

hfujita: And it looks like `sys.exit(msg)` does seem to emit `msg` to _stderr_, which is good.

gfx (approved):
👏

PR #509, opened 2022-03-18
qlog-adapter: start reading from idx == 0

hfujita:
This fixes an issue that qlog-adapter ignored the first event (idx == 0) when gathering sent frame information.


gfx (approved):
LGTM!

PR #510, opened 2022-03-18
qlog-adapter: Add length to packet send/recv event

hfujita:
According to the [qlog draft](https://quicwg.org/qlog/draft-ietf-quic-qlog-quic-events.html#name-packetheader), packet header can include `length` field but qlog-adapter currently does not emit it.


gfx (approved):
LGTM!

PR #511, opened 2022-03-22
qlog-adapter: Add handler for `max_streams_receive`

hfujita:
The `receive` handler is missing while the `send` version exists. I assume this is due to an error.


gfx (approved):
LGTM!

PR #529, opened 2022-10-13
[build] fix build on Ubuntu 16.04 -- older cmake might require empty vars to be quoted

gfx:
🙏 

---

I wanted to add an Ubuntu 16.04 scenario to CI, but because our Ubuntu 16.04 image does not have `netstat(1)` it needed some tricks to write portable UDP socket checking in e2e.t. I'll make a separate PR for such CI scenarios.


t/e2e.t
```diff
@@ -347,7 +347,7 @@ sub spawn_server {
         exec @cmd;
         die "failed to exec $cmd[0]:$?";
     }
-    while (`netstat -na` !~ /^udp.*\s127\.0\.0\.1[\.:]$port\s/m) {
+    while (`ss --udp --all src "127.0.0.1:$port"` !~ /^\w+\s+\d+\s+\d+\s+127\.0\.0\.1:$port\s/m) {
```
gfx, line 350: Instead of installing netstat(1) to our Ubuntu 16.04 image, this uses ss(1), which is installed by default.

PR #664, opened 2026-04-17
remote_cid: retired CID shift copies too few survivors

jedisct1:
`quicly_remote_cid_shift_retired` removes `count` retired connection IDs from the front of the queue, decrements `set->retired.count`, and then shifts survivors forward.

The existing loop copies only `count` elements from `i + count` to `i`, which is insufficient whenever more than `count` entries survive.

As a result, the surviving prefix is corrupted and later retirement retries can reference the wrong sequence numbers.


kazuho:
Thank you for the fix!

PR #671, opened 2026-07-15
Discard duplicate packet numbers before frame processing

### LINKED ISSUE #660: Duplicated Packet Numbers
### Tested Build

https://github.com/h2o/quicly/commit/fce6a5f2eae1786320953d049b5077c9fa7fd331

### Observation

We observed that quicly appears to decrypt and process packets even when a packet with the same packet number in the same packet number space was already processed. In our tests, replaying an already accepted packet results in the packet being decrypted again and the frames being processed again. While the logical outcome is idempotent, this behavior can still impose unnecessary CPU work. Is there a reason quicly is implemented this way?

### Expected Behavior 

RFC 9000, Section 12.3 states:
> A receiver MUST discard a newly unprotected packet unless it is certain that it has not processed another packet with the same packet number from the same packet number space. Duplicate suppression MUST happen after removing packet protection for the reasons described in [Section 9.5](https://www.rfc-editor.org/rfc/rfc9001#section-9.5) of [[QUIC-TLS](https://www.rfc-editor.org/rfc/rfc9000.html#QUIC-TLS)].

fidgetaryan445:
Fixes #660.

Every packet(even duplicates) was being initially handled by `handle_payload()` . Added a function `quicly_range_exist()` which checks if the packet number is already covered in one of the continuous packet streams that was received.

### DIFF
--- a/lib/quicly.c
+++ b/lib/quicly.c
@@ -7394,7 +7394,7 @@ static quicly_error_t do_receive(quicly_conn_t *conn, struct sockaddr *dest_addr
     size_t epoch, path_index;
     ptls_iovec_t payload;
     uint64_t pn, offending_frame_type = QUICLY_FRAME_TYPE_PADDING;
-    int is_ack_only, is_probe_only;
+    int is_ack_only, is_probe_only, is_duplicate;
     quicly_error_t ret;
 
     assert(src_addr->sa_family == AF_INET || src_addr->sa_family == AF_INET6);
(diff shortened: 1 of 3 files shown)


lib/quicly.c
```diff
@@ -7600,6 +7600,9 @@ static quicly_error_t do_receive(quicly_conn_t *conn, struct sockaddr *dest_addr
         goto Exit;
     }
 
+    /* duplicate check must happen after decryption; see RFC 9000 12.3, RFC 9001 9.5 */
+    is_duplicate = quicly_ranges_exists(&(*space)->ack_queue, pn);
```
kazuho, line 7604: Memo to myself:
* We should optimize for the ordinal case; i.e., pn was no less than next_expected_packet_number.
* If the PN was below any value stored in the ack queue, we need to ingore (S12.3)

fidgetaryan445: If you want me to make any further contribution to current PR , I am all in to help

PR #677, opened 2026-08-10
[pico] impleent ABE

kazuho:
This PR implements RFC 8511 (TCP Alternative Backoff with ECN) for pico; the beta chosen for CE is 0.85, as we use 0.7 for packet losses.

The PR also adds -F (flow isolation) and -A (AQM algorithm) options to the simulator; it can simulate fq_codel with `-F -A codel`.
As to the AQM algorithm, currently supported ones are: `codel` and `step` (mark everything when the queue is above threshold).

The chart below is the interaction with Rapid Start; first congestion is reported via a packet loss, because Codel marks packets only if the queue was occupied for longer than 100ms, which is much longer than the path RTT of a typical DSL connection. However, once exiting startup, congestions are reported via CE marks, and there is no retransmits (look at the blue line).
Screenshot 2026-08-11 at 7 27 25

The 2nd chart shows the bandwidth occupation on a DSL network with fq_codel. Pico with ECN turned on gets slightly more bandwidth than that with ECN turned off; this difference is due to the different betas (0.85 on CE vs 0.7 on congestion). The RTT of the simulated network is 30ms, and codel reports congestion (either via CE or drop) when the buffer is >=5ms. If codel drops a packet and CWND is reduced to 0.7x of the original, the sender cannot fulfill the bottleneck (i.e., `(30 + 5) * 0.7 = 24.5 < 30`). Conversely, when codel marks a packet, CWND is reduced only to 0.85x, and the sender continues to fulfill the bottleneck (`(30 + 5) * 0.85 = 29.75 ~ 30`).
Screenshot 2026-08-11 at 7 27 37

### DIFF
--- a/include/quicly/cc.h
+++ b/include/quicly/cc.h
@@ -540,7 +540,9 @@ inline void quicly_cc_rapid_start_on_recovery(struct st_quicly_cc_rapid_start_t 
 #undef ENTRY
     };
 
-    *cwnd -= factors[rs->recovery.by_ecn].ack * bytes_acked + factors[rs->recovery.by_ecn].loss * bytes_lost;
+    uint32_t reduction = factors[rs->recovery.by_ecn].ack * bytes_acked + factors[rs->recovery.by_ecn].loss * bytes_lost;
+    assert(reduction <= *cwnd && "CWND never underflows");
+    *cwnd -= reduction;
     if (*cwnd < rs->recovery.cwnd_floor)
         *cwnd = rs->recovery.cwnd_floor;
 }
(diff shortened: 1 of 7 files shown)


kazuho:
For comparison, the chart below is the behavior of quicly without ECN on the same path. Within the observed window of 2 seconds:
* There are multiple recovery events triggered by AQM-triggered losses. Each recovery event delays delivery of data to the application until losses are repaired.
* In stable condition, queue remains empty for ~40% of time between each recovery event, leading to underutilization of of the bottleneck. This is because Codel marks a packet before the queue becomes deep enough to compensate for the 0.7x reduction when the idle RTT is 30ms.
* Total bytes delivered after 2 seconds is ~3% lower without ECN (6.87MB vs. 7.10MB).

Screenshot 2026-08-12 at 15 31 09

This is a good example of how AQM could hurt the throughput of non-ECN flows.

PR #678, opened 2026-08-11
implement Cuback: an ack-driven Cubic / re-implement Reno and Cubic; all in cc-pico.c

kazuho:
This PR has subsumed the following PRs:
* #683
* #686

Below is about Cuback.

## The Design

Quoting from the doc-comment of `cuback_cwnd_to_bytes_sent`:

Cuback is an ACK-driven variant of Cubic. It traces the same two curves as Cubic - W_cubic and the Reno-friendly W_reno shown below - but drives them by the bytes being acked rather than by the clock.

Let Tr be the time at which the Reno-friendly curve reaches cwnd_prior:
```
   Tr = (cwnd_prior^2 - Wepoch^2) / (2 * alpha * MSS * bandwidth)
```

The two time-domain curves are:
```
  W_cubic(t) = C * MSS * (t - K)^3 + Wmax
  W_reno(t)  = sqrt(Wepoch^2 + 2 * alpha * MSS * bandwidth * t),             t <= Tr
               sqrt(cwnd_prior^2 + 2 * MSS * bandwidth * (t - Tr)),          t > Tr
```

As both curves are monotonically increasing, CWND - being their maximum - is monotonically increasing as well. Therefore, the amount sent can be recovered from CWND, the inverse of a maximum being the minimum of the inverses. Inverting the curves gives the time at which each reaches `w`. Multiplying that time by bandwidth converts the result to bytes sent since the epoch began:
```
  bytes_cubic(w) = bandwidth * (K + cbrt((w - Wmax) / (C * MSS)))
  bytes_reno(w)  = (w^2 - Wepoch^2) / (2 * alpha * MSS),                     w <= Wmax
                   (Wmax^2 - Wepoch^2) / (2 * alpha * MSS)
                     + (w^2 - Wmax^2) / (2 * MSS),                           w > Wmax
  bytes_sent(w)  = min(bytes_cubic(w), bytes_reno(w))
```

Wepoch is equal to ssthresh, therefore the only Cuback-specific states that need to be retained are:
* Wmax
* bandwidth observed at the last congestion event
* (cwnd_prior as a value separate from Wmax, if fast convergence is needed)
* (alpha, if ABE is supported and uses a different alpha when reacting to ECN-CE)

These values remain immutable during each epoch. The congestion-avoidance trajectory is expressed entirely as pure functions using them, taking CWND as the only parameter.

Using the `bytes_sent` function, the congestion controller calculates bytes needed to be acked before incrementing the CWND (`bytes_sent(cwnd + mtu) - bytes_sent(cwnd)`) and drives congestion avoidance, the same way as in the case of Reno.

In addition to being simple, the approach has two positive side-effects:
* As CWND is driven by acks rather than by time, CWND does not advance while nothing is inflight. Congestion control becomes safer without accurate app-limited tracking.
* Because the clock is the ack stream, a flow whose share is being taken advances more slowly than one that is taking, thereby yielding bandwidth. Fast convergence could be still helpful.

## Benchmark

The numbers were collected using [t/simulator.c](https://github.com/h2o/quicly/blob/master/t/simulator.c) at c7a8294, using the conditions defined in [t/simulator-plot.rb](https://github.com/h2o/quicly/blob/master/t/simulator-plot.rb). Each percentage shows the bandwidth allocated to the flow, when cubic is competing on the other side.

||DSL|LTE|5G|
|:--:|:--:|:--:|:--:|
|cubic (sanity)|50.0%|51.0%|49.9%|
|pico (old; 8283105)|57.3%|32.9%|30.6%|
|pico (new; this branch)|57.2%|34.0%|32.2%|
|cuback|50.1%|50.4%|49.4%|

Pico is our congestion controller that uses linear increase but reaches Wmax equally fast as Cubic does. However, it underperforms  on high BDP networks when competing against Cubic, even with #679 adopted by the current branch, due to the following reasons:
* Pico uses linear increase. Total amount of bytes sent is the integral of the CWND increase curve. The integral of linear increase is about 10% less than that of the cubic curve.
* Early in the epoch, more than half of the packets in flight belong to Cubic, due to its rapid increase. That slows down the ack-driven clock of Pico, thereby yielding bandwidth.

On small BDP path (DSL), Pico is more aggressive than Cubic, and that is because its alpha_cubic is fixed to 1 (rather than a fraction until CWND reaches cwnd_prior).

Cuback resolves both issues by using the same cubic curve as Cubic. As can be seen, it is in the margin of error against Cubic.


kazuho:
Both CUBIC and Cuback behaved differently from the Reno-friendly behavior defined by RFC 9438, with CUBIC being substantially more aggressive than Cuback. The commits since af44c7e align both implementations with the RFC. In the simulator benchmark, their throughput relative to Reno has consequently become comparable.

The fixes are:

- af44c7e changes Cuback’s Reno-friendly inverse from the continuous integral to the packet-granular Reno recurrence, scheduling each one-MSS increase after the correct number of acknowledged bytes.
- 4faf843 makes CUBIC expose its continuous `W_est` in whole-MSS steps.
- 16a3412 aligns their epoch origins with the beginning of congestion avoidance: CUBIC no longer starts its clock at congestion detection, and Cuback no longer receives one CWND of post-recovery ACK credit.

To compare their Reno-friendly growth directly, I ran `t/simulator.c` with both controllers forced onto the Reno-friendly branch. The current results come from 16a3412. For the ef3c61b control, Cuback’s credit was disabled so that both sides start from the same origin and the benchmark isolates the growth curves.

Benchmark conditions:

- 1 MB/s bottleneck
- 5 ms one-way delay
- 20 ms queue
- pacing and Rapid Start enabled
- newcomer starts at 2 seconds
- 60-second simulation
- 30 runs per cell
- throughput measured from contiguously delivered bytes at the receiver
- results normalized against a Reno newcomer

With one Reno incumbent:

| Policy | Before fixes (ef3c61b, no-credit control) | Current (16a3412) |
|---|---:|---:|
| CUBIC | +27.65% (625,029 ± 24,260 B/s) | +15.32% (565,261 ± 11,584 B/s) |
| Cuback | +16.01% (568,016 ± 9,306 B/s) | +16.93% (573,162 ± 10,338 B/s) |

Reno baselines: 489,635 ± 6,090 B/s before and 490,186 ± 7,503 B/s currently.

With four Reno incumbents:

| Policy | Before fixes (ef3c61b, no-credit control) | Current (16a3412) |
|---|---:|---:|
| CUBIC | +47.49% (289,965 ± 7,732 B/s) | +26.20% (247,133 ± 8,616 B/s) |
| Cuback | +16.56% (229,162 ± 8,910 B/s) | +28.57% (251,774 ± 8,766 B/s) |

Reno baselines: 196,604 ± 5,723 B/s before and 195,825 ± 5,314 B/s currently.

The CUBIC–Cuback difference fell from 11.64 to 1.61 percentage points relative to Reno with one incumbent, and from 30.93 to 2.37 points with four incumbents.

kazuho:
A follow-up on the benchmark section: although Cuback and CUBIC converge to similar allocations over long runs, Cuback can acquire bandwidth materially faster when joining an established CUBIC flow. This matters for short flows, which may finish before equilibrium is reached.

### Newcomer benchmark

The benchmark uses Quicly commit `14194a71610049f3ee31c5acee626cd0d2f51f0b` with:

- CUBIC incumbent: `-c cubic -i 30 -p`
- Newcomer: either CUBIC or Cuback, both using `-i 30 -j 60 -p -R`
- 100 matched seed/arrival-phase pairs per policy
- newcomer throughput measured from bytes dequeued by the bottleneck
- no failed or low-utilization runs excluded

The full newcomer trajectories are here:

- [5G SVG](https://github.com/h2o/quicly/blob/experiment/cuback-vs-cubic-newcomer/results/cuback-vs-cubic-newcomer-5g.svg?raw=true)
- [DSL SVG](https://github.com/h2o/quicly/blob/experiment/cuback-vs-cubic-newcomer/results/cuback-vs-cubic-newcomer-dsl.svg?raw=true)
- [LTE SVG](https://github.com/h2o/quicly/blob/experiment/cuback-vs-cubic-newcomer/results/cuback-vs-cubic-newcomer-lte.svg?raw=true)
- [Half DSL SVG](https://github.com/h2o/quicly/blob/experiment/cuback-vs-cubic-newcomer/results/cuback-vs-cubic-newcomer-half-dsl.svg?raw=true)
- [Quarter DSL SVG](https://github.com/h2o/quicly/blob/experiment/cuback-vs-cubic-newcomer/results/cuback-vs-cubic-newcomer-quarter-dsl.svg?raw=true)
- [Eighth DSL SVG](https://github.com/h2o/quicly/blob/experiment/cuback-vs-cubic-newcomer/results/cuback-vs-cubic-newcomer-eighth-dsl.svg?raw=true)

To avoid defining “early” in terms of policy- and network-dependent congestion-avoidance periods, I reran the benchmark using ten fixed one-second windows after newcomer arrival. Each entry below is `CUBIC → Cuback` in Mbit/s, followed by Cuback’s relative difference. Each throughput value is the arithmetic mean across the 100 runs. Confidence intervals are paired-bootstrap intervals for the ratio of means.

| Second | 5G | DSL | LTE | Half DSL | Quarter DSL | Eighth DSL |
|---:|---:|---:|---:|---:|---:|---:|
| 1 | 14.646 → 14.902 (+1.7%) | 7.504 → 8.390 (+11.8%) | 3.956 → 3.987 (+0.8%) | 5.207 → 5.423 (+4.2%) | 3.635 → 3.708 (+2.0%) | 2.130 → 2.134 (+0.2%) |
| 2 | 24.673 → 27.526 (+11.6%) | 9.941 → 11.124 (+11.9%) | 7.703 → 8.455 (+9.8%) | 6.242 → 6.562 (+5.1%) | 3.915 → 3.853 (−1.6%) | 2.026 → 2.049 (+1.1%) |
| 3 | 28.425 → 31.709 (+11.6%) | 10.412 → 11.483 (+10.3%) | 8.902 → 9.956 (+11.8%) | 6.437 → 6.688 (+3.9%) | 3.830 → 3.737 (−2.4%) | 1.957 → 1.939 (−0.9%) |
| 4 | 29.876 → 32.640 (+9.3%) | 11.021 → 11.924 (+8.2%) | 9.345 → 10.186 (+9.0%) | 6.554 → 6.975 (+6.4%) | 3.788 → 3.721 (−1.8%) | 1.900 → 1.900 (−0.0%) |
| 5 | 30.458 → 32.781 (+7.6%) | 11.785 → 12.237 (+3.8%) | 9.551 → 10.331 (+8.2%) | 6.734 → 7.159 (+6.3%) | 3.797 → 3.772 (−0.7%) | 1.856 → 1.873 (+0.9%) |
| 6 | 30.255 → 32.662 (+8.0%) | 12.030 → 12.418 (+3.2%) | 9.330 → 10.020 (+7.4%) | 6.894 → 7.173 (+4.1%) | 3.744 → 3.831 (+2.3%) | 1.835 → 1.899 (+3.5%) |
| 7 | 30.246 → 32.598 (+7.8%) | 12.056 → 12.480 (+3.5%) | 9.513 → 10.215 (+7.4%) | 7.031 → 7.224 (+2.7%) | 3.743 → 3.825 (+2.2%) | 1.844 → 1.875 (+1.7%) |
| 8 | 30.117 → 32.083 (+6.5%) | 12.030 → 12.745 (+5.9%) | 9.246 → 10.069 (+8.9%) | 7.129 → 7.308 (+2.5%) | 3.715 → 3.836 (+3.3%) | 1.872 → 1.858 (−0.7%) |
| 9 | 30.000 → 32.601 (+8.7%) | 12.330 → 13.062 (+5.9%) | 9.628 → 10.253 (+6.5%) | 7.157 → 7.321 (+2.3%) | 3.761 → 3.862 (+2.7%) | 1.893 → 1.849 (−2.3%) |
| 10 | 30.164 → 32.566 (+8.0%) | 12.806 → 13.396 (+4.6%) | 9.648 → 10.573 (+9.6%) | 7.194 → 7.327 (+1.8%) | 3.757 → 3.811 (+1.4%) | 1.892 → 1.907 (+0.8%) |

Paired-bootstrap confidence intervals put the Cuback gain above zero in every one-second window on 5G and LTE. DSL is also clearly positive in most windows. Half DSL has a smaller early benefit, while Quarter and Eighth DSL show no consistent difference.

The startup algorithms themselves behave identically: measured from arrival to the first recovery entry, CUBIC and Cuback have identical throughput. The difference begins after the first congestion event.

### ACK-clock ablation

I then added a hybrid that retains CUBIC’s startup, recovery, Wmax, fast convergence, cubic target integration, and `W_est`, changing only the cubic curve’s clock from wall time to acknowledged-byte-derived virtual time:

```c
ack_clock_time += bytes_acked / bandwidth_at_congestion;
```

All newcomers again use `-i 30 -j 60 -p -R`.

[ACK-clock ablation SVG](https://github.com/h2o/quicly/blob/experiment/cuback-ackclock-ablation/results/lte-newcomer-ackclock-ablation.svg?raw=true)

| Horizon | CUBIC | ACK-clock CUBIC | Cuback | Clock effect as share of total gain |
|---:|---:|---:|---:|---:|
| 5 s | 0.2601 | 0.2802 | 0.2834 | 87% |
| 10 s | 0.2889 | 0.3118 | 0.3128 | 96% |

The ACK-clock hybrid beat ordinary CUBIC in all 100 paired runs at both five and ten seconds. This shows that replacing the wall clock accounts for nearly all of Cuback’s early newcomer gain; the different Reno-friendly estimator is a comparatively small residual.

### Direct mechanism measurement

The ablation above establishes the clock-source effect. To observe the mechanism directly, I also measured the ACK-clock hybrid at recovery entry from each of its first ten CA periods.

For the observed pre-reduction window `Wloss`, define:

```text
X = K + cbrt((Wloss - Wmax) / (C * MSS))
Y = recovery_time - CA_epoch_start
```

`X` is when ordinary wall-clock CUBIC’s cubic curve says the observed CWND should have been reached. `Y` is how much wall time actually elapsed. Thus `X/Y > 1` means the ACK-clock flow reached that CWND faster than wall-clock CUBIC predicts.

For CA1, across all 100 runs:

| Measurement | Result |
|---|---:|
| Nominal CUBIC time `X` | 2.529 s [2.455, 2.588] |
| Actual wall time `Y` | 1.623 s [1.541, 1.654] |
| `X/Y` | 1.546× [1.514, 1.595] |
| Cubic curve dominant | 100/100 |
| Fast convergence active | 0/100 |

The later periods have `X/Y` close to one. In other words, the virtual ACK clock advances the cubic curve about 1.55 times as fast as wall time during the first CA period, before settling to approximately wall-clock pace.

This also rules out fast convergence and `W_est` as explanations for the initial result: neither controls CA1 in these runs. Together, the throughput benchmark, isolated clock ablation, and recovery-entry measurement indicate that Cuback’s early newcomer advantage comes from using acknowledged bytes—rather than wall time—to drive the cubic curve.

Full ablation methodology and results are in [`experiment/cuback-ackclock-ablation`](https://github.com/h2o/quicly/tree/experiment/cuback-ackclock-ablation).

PR #683, opened 2026-08-15
merge reno, cubic into cc-pico.c

kazuho:
Builds on top of #678.

Until #678, each CC policy (reno, pico, cubic) were implemented in different files. That led to new features (e.g., Rapid Start, ECN) added only to pico, leaving others behind.

#678 implements cuback as a variation of pico in cc-pico.c, and the added complexity was very small. Hence it is natural to ask: could we also implement other policies as variations in cc-pico.c?

This pull request does that. We do have more if..else statements dealing with different policies, but it seems this is more manageable than having one .c file for each policy.

Specifically, this PR changes the following:

* Cubic is also implemented in cc-pico.c.
  * Old implementation (cc-cubic.c) is preserved, but is renamed to "cubic-legacy."
  * The new Cubic controller nicely integrates Rapid Start and ABE.
  * It also implements cc-limited properly.
* Reno is moved into cc-pico.c.
  * The old implementation is removed.
  * It now uses beta of 0.5.
  * Rapid Start can be enabled, but as Rapid Start assumes a beta of 0.7, transition of CA might have issues, though they are likely to be minor.


kazuho:
Merging into #678, as this is essentially becoming a one big change.

PR #689, opened 2026-09-05
Accelerated Bottleneck Bandwidth Adaptation (ABBA) for CUBIC and Cuback

### LINKED ISSUE #687: Adding random-loss resilience to loss-based congestion controllers
__Below is the design note:__

The objective is to avoid reducing the congestion window for isolated non-congestive losses while remaining responsive to tail-drop queues and AQMs.

To achieve this objective, an RTT estimator called `fqRTT` tracks the RTT samples obtained when the queue is estimated full. The estimator maintains a smoothed RTT, `sfqRTT`, and an RTT variation, `fqRTTvar`.

For each lost packet, `latestRTT`—the RTT sample obtained from the ACK that caused the packet to be declared lost—is used as the closest available proxy for the path delay around the time of the drop. The classifier then acts as follows:

* If `fqRTT` has not received a sample, declare congestion.
* Otherwise, if this is the second packet reported lost within one round trip, declare congestion.
* Otherwise, if `latestRTT >= min(sfqRTT - fqRTTvar, minRTT + 5ms)`, declare congestion.
* Otherwise, treat the loss as non-congestive and do not reduce the congestion window.

Upon exiting recovery, `fqRTT` is updated if it has not previously received a sample or if the loss episode contained multiple lost packets. One sample is taken using the minimum `latestRTT` associated with the episode, including any preceding loss that contributed to the multi-loss condition.

The 5 ms guard is important for AQMs such as CoDel, which may deliberately signal congestion by dropping only one packet near the bottom of the queue.

TODO: Discuss the initial value of `fqRTT` (it can be 5ms assuming that most bottleneck queues are 10ms or deeper).

kazuho:
## Problem

Radio paths violate an assumption CUBIC depends on: that available bandwidth is roughly stable. Radio quality varies, and handover between stations or satellites moves capacity in steps. When capacity drops the queue overflows; when it recovers the flow is left well under the pipe, and CUBIC's cubic-curve recovery is too slow to climb back before the next disturbance.

What's needed is to increase CWND faster whenever the flow is demonstrably below the pipe.

The same condition — an empty queue while the flow is not the constraint — is what random loss also produces, so the mechanism additionally mitigates underutilization on lossy paths. On a lossy Corporate_WiFi profile CUBIC holds **46.7%** of link goodput at 5% loss-per-RTT and **34.3%** at 10%.

## Approach

ABBA combines two ideas.

**1. Increase rapidly while RTT sits at the floor.** When the latest RTT is at its minimum and being driven down, the bottleneck queue is empty and the flow is not the constraint. ABBA adds a second growth term alongside the CUBIC/Cuback curve and takes the larger of the two, at `max(2ms / rtt_threshold, 2.5%)` per RTT — about 2 ms of additional queue per round. RTT feedback lags a round, so this self-limits near 4.5 ms of standing queue and stops. The threshold adapts: `(minRTT + previous_period_min) / 2`, floored at `minRTT + 2ms`, capped at `current_period_min + 2ms`.

**2. Gate on confirmed queue existence.** Acceleration requires prior evidence that this path *has* a queue to refill: `full_rtt` — the smoothed RTT captured when the first recovery exits — must exceed `minRTT + 10ms` and sit more than 5% above the latest RTT.

The gate is not only a safety guard. Because `full_rtt` is a live estimate of what a full queue looks like, it also drives the second recovery mechanism: if smoothed RTT stays below `(minRTT + full_rtt) / 2` for twice the time a non-losing CUBIC flow would need to rebuild the queue, the path characteristics may have changed. The controller retains CWND, discards the CA trajectory, and re-enters slow start to re-learn `full_rtt`.

That second path is what carries the design under frequent loss, where the flow never rebuilds a queue on its own and the RTT gate alone has nothing to work with. It is directly visible in the policy comparison: LTE CUBIC standalone at 10% loss/RTT reaches **90.6%** with recalibration enabled and **80.6%** without.

ECN-CE counts as a high-queue observation, including marks received inside a recovery period.

## Safety: increase-side only

**ABBA never suppresses, defers, or scales a congestion response.** Every loss and every CE mark produces the same multiplicative decrease as stock CUBIC/Cuback. `beta` is derived solely from the controller type and whether the signal was ECN; nothing on the reduction path is conditioned on ABBA state or policy.

This is deliberately not a random-loss classifier. A classifier decides a loss was non-congestive and withholds the reduction, which stakes correctness on the classification being right — and it is wrong precisely when real congestion resembles random loss. ABBA declines that trade: it buys recovery entirely on the increase side and pays full price on every signal. The consequence is that a misfiring RTT gate on some path nobody simulated costs one round of over-growth followed by a normal decrease, rather than a missed congestion response. The bound holds by construction, not by calibration.

The same invariant covers the capacity *drop* described above: the reduction path is untouched, so the resulting queue overflow produces the standard response, with ABBA contributing at most its ~4.5 ms of extra queue at the moment of the drop.

Recalibration re-enters slow start from a large CWND, which needs its own argument. It is gated on the *absence* of queue pressure from any source: `last_high_queue_at` is refreshed whenever smoothed RTT reaches the midpoint between `minRTT` and `full_rtt`, and whenever a CE mark arrives, so the timer matures only after sustained evidence that neither the flow itself nor any competitor is queueing at the bottleneck. The wait is measured in CUBIC's own units — `high_rtt_interval` is how long a flow at this window scale takes to climb back from a reduction and refill the queue — and recalibration needs twice that to pass untouched. CUBIC would have filled the pipe twice over; nothing did.

## Design criteria and results

Simulator, 18 network profiles × {tail-drop, CoDel}. Loss probability is set so a full pipe sees at least one loss per RTT with the stated probability. Standalone: 7×100 s runs measuring the last 50 s. Competition: 7×160 s per flow order measuring the last 80 s. Self-competition: 5×320 s, final 40 s window. Full tables are in the commit messages of `2766165` (0x2) and `2a2cde1` (0x1).

**On par with stock when the network is stable and lossless.** Clean-path competition against stock sits at parity on every profile; the widest deviations are Corporate_WiFi 58.1/44.1 and WiFi@Office 53.3/45.3, both inside the stock-vs-stock baseline spread. On the shallow synthetic queues the gate correctly keeps acceleration off entirely: `queue5ms` 49.6/49.6, `queue2.5ms` 51.0/50.8.

**Recovers quickly under random loss** (standalone goodput, stock → ABBA, tail-drop):

| Profile | CUBIC 5% | CUBIC 10% | Cuback 5% | Cuback 10% |
|---|---|---|---|---|
| Corporate_WiFi | 46.7 → **97.5** | 34.3 → **97.1** | 55.1 → **97.6** | 34.4 → **97.3** |
| WiFi@Office | 72.5 → **97.7** | 51.6 → **97.7** | 72.4 → **97.7** | 50.0 → **97.7** |
| 5G | 60.6 → **95.0** | 39.9 → **90.7** | 64.7 → **96.7** | 43.3 → **93.5** |
| Home_WiFi | 51.7 → **93.8** | 36.5 → **88.4** | 51.9 → **95.6** | 35.9 → **91.0** |
| LTE | 89.0 → **95.1** | 62.9 → **90.6** | 89.6 → **97.3** | 67.6 → **94.4** |
| LEO | 80.1 → **95.8** | 53.4 → **91.3** | 75.3 → **97.5** | 56.9 → **94.7** |

**Stays below a competing stock flow that is not experiencing loss.** This was a design criterion, and it is met with margin: a lossy ABBA flow against a clean stock flow takes **17–49%** across all profiles — never parity. Ten clean CUBIC flows against one ABBA flow on DSL (15 repetitions, mean of flow-order medians): the ABBA flow takes **10.21%** with no loss and **9.62%** at 5% loss/RTT, against a 10% fair share.

**Self-converges.** When loss is rare the gates stay shut and behavior reduces to stock, which converges; acceleration is additionally self-limiting, since a flow that opens its gate builds queue and closes it again. When loss is frequent, recalibration is the convergence driver — both flows re-probe symmetrically. Self-competition at 5% loss gives medians near 50% with 85–98% utilization across all profiles.

## Policy flags

`quicly_context_t::accel_adaptation` replaces the earlier boolean. **Default is 0 (off).**

| Flag | Value | Behavior |
|---|---|---|
| `QUICLY_CC_ACCEL_ADAPTATION_INCREASE_ALWAYS` | `0x1` | Bypasses the `full_rtt` guard; adaptive RTT gate still applies |
| `QUICLY_CC_ACCEL_ADAPTATION_RECALIBRATE` | `0x2` | Observes `full_rtt`, enables recalibration, and (unless `0x1`) uses `full_rtt` as the increase guard |
| `QUICLY_CC_ACCEL_ADAPTATION_ON` | `= 0x2` | Recommended setting; what simulator `-T` selects |

`full_rtt` and high-queue observations are maintained regardless of policy; the flags govern only how they are consumed. `0x1` drops both roles of the gate at once — it is weaker standalone under heavy loss (LTE CUBIC 80.6% vs 90.6% at 10%/RTT, for the reason above) and on the synthetic `queue5ms`/`queue2.5ms` profiles takes 63–77% share against stock. Those profiles are stress cases rather than representative access-network queues, but they show what the gate is buying.

## Open items

- All results are from the simulator, which is not a real network. Its bottleneck rate is fixed, so the bandwidth-step case that motivates the design is argued from mechanism rather than demonstrated; beyond that it does not model correlated radio loss, ACK aggregation, or real cross traffic. Real-network validation is pending.
- `0x3` has unit coverage but no benchmark run.
- `-T` takes no argument, so `0x1`/`0x3` aren't selectable from the simulator; the `0x1` figures come from a separate always-increase build.
- Competition between *lossy* ABBA flows at different RTTs is untested. The increase rate targets an absolute queue depth (2 ms per round), so in wall-clock terms a short-RTT flow builds queue faster than a long-RTT one; whether this compounds CUBIC's existing RTT bias is an open question. Flows not experiencing non-congestive loss are unaffected, the gates keeping them on the stock trajectory.
- If the path's idle RTT rises during a connection, ABBA as implemented becomes less effective: the increase gate cannot rise above `(rtt->minimum + min_rtt_previous_period) / 2`, and `rtt->minimum` only ever moves down.
- All competition data is against stock CUBIC/Cuback. Behavior against BBR — a realistic peer on these paths, and one that deliberately drains the queue — is untested.

Closes #687.

### DIFF
--- a/include/quicly.h
+++ b/include/quicly.h
@@ -156,7 +156,7 @@ QUICLY_CALLBACK_TYPE(quicly_error_t, generate_resumption_token, quicly_conn_t *c
  * called to initialize a congestion controller for a new connection.
  * should in turn call one of the quicly_cc_*_init functions from cc.h with customized parameters.
  */
-QUICLY_CALLBACK_TYPE(void, init_cc, quicly_cc_t *cc, uint32_t initcwnd, int normalize_mtu, int random_loss_tolerance, int64_t now);
+QUICLY_CALLBACK_TYPE(void, init_cc, quicly_cc_t *cc, uint32_t initcwnd, int normalize_mtu, unsigned accel_adaptation, int64_t now);
 /**
  * reference counting.
  * delta must be either 1 or -1.
@@ -390,9 +390,9 @@ struct st_quicly_context_t {
      */
     unsigned normalize_cc_mtu : 1;
     /**
-     * if congestion control should tolerate random loss
+     * controls accelerated bottleneck bandwidth adaptation; see `QUICLY_CC_ACCEL_ADAPTATION_*`
      */
-    unsigned random_loss_tolerance : 1;
+    unsigned accel_adaptation : 2;
     /**
      *
      */
--- a/include/quicly.h
+++ b/include/quicly.h
@@ -156,7 +156,7 @@ QUICLY_CALLBACK_TYPE(quicly_error_t, generate_resumption_token, quicly_conn_t *c
  * called to initialize a congestion controller for a new connection.
  * should in turn call one of the quicly_cc_*_init functions from cc.h with customized parameters.
  */
-QUICLY_CALLBACK_TYPE(void, init_cc, quicly_cc_t *cc, uint32_t initcwnd, int normalize_mtu, int64_t now);
+QUICLY_CALLBACK_TYPE(void, init_cc, quicly_cc_t *cc, uint32_t initcwnd, int normalize_mtu, int random_loss_tolerance, int64_t now);
 /**
  * reference counting.
  * delta must be either 1 or -1.
@@ -389,6 +389,10 @@ struct st_quicly_context_t {
      * the default contexts
      */
     unsigned normalize_cc_mtu : 1;
+    /**
+     * if congestion control should tolerate random loss
+     */
+    unsigned random_loss_tolerance : 1;
     /**
      *
      */
--- a/include/quicly/cc.h
+++ b/include/quicly/cc.h
@@ -140,9 +140,10 @@ struct st_quicly_cc_accelerated_increase_t {
      */
     uint32_t min_rtt_previous_period;
     /**
-     * Latest time at which the smoothed RTT was at least halfway between the minimum RTT and `full_rtt`.
+     * Latest time at which a high queue was indicated by ECN-CE or by the smoothed RTT reaching halfway between the minimum RTT
+     * and `full_rtt`.
      */
-    int64_t last_high_rtt_at;
+    int64_t last_high_queue_at;
     /**
      * Expected time in milliseconds for the active congestion-avoidance trajectory to produce a high-RTT observation.
      */
--- a/include/quicly/cc.h
+++ b/include/quicly/cc.h
@@ -124,6 +124,28 @@ struct st_quicly_cc_cuback_t {
 };
 
 /**
+ * State used to accelerate CWND increase after Random Loss Tolerance is enabled.
+ */
+struct st_quicly_cc_accelerated_increase_t {
+    /**
+     * Queue-overflow RTT observed when calibration slow start encountered a packet loss, or zero before an observation is obtained.
+     */
+    uint32_t qortt;
+    /**
+     * Latest CWND cap established for accelerated increase, or zero when no cap is available.
+     */
+    uint32_t target_cwnd;
+    /**
+     * Beginning of the current path-change observation interval, or zero when no interval is active.
+     */
+    int64_t observation_start;
+    /**
+     * Maximum smoothed RTT observed during the current path-change observation interval.
+     */
+    float max_smoothed_rtt;
+};
+
+/**
  * State used by the Cubic policy implemented by cc-pico.c; see `quicly_cc_type_cubic`.
  */
 struct st_quicly_cc_cubic_t {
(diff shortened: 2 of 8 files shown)


kazuho:
> * If the path's idle RTT rises during a connection, ABBA as implemented becomes less effective: the increase gate cannot rise above `(rtt->minimum + min_rtt_previous_period) / 2`, and `rtt->minimum` only ever moves down.

FWIW, as to this point, I think tracking the EMEA of per-epoch minimum and using its `(smoothed - variance / 2) + 2ms` as a gate could make sense, though I'm not sure how much that'd be an improvement for HTTP when most connections are short-lived. I'm inclined to test this as-is in prod first, and then try the EMEA variant using the same traffic to _confirm that EMEA is no more aggressive_.

kazuho:
Closing in favor of #698. We might resurrect some good parts from this PR.

PR #690, opened 2026-09-05
address compliance issues

sleepybishop:
This pr addresses the findings from https://github.com/h2o/h2o/issues/3607.

The picotls related items are deferred until https://github.com/h2o/picotls/pull/617 is merged.

### DIFF
--- a/include/quicly/cc.h
+++ b/include/quicly/cc.h
@@ -347,7 +347,7 @@ struct st_quicly_cc_type_t {
     /**
      * Called when persistent congestion is observed.
      */
-    void (*cc_on_persistent_congestion)(quicly_cc_t *cc, const quicly_loss_t *loss, int64_t now);
+    void (*cc_on_persistent_congestion)(quicly_cc_t *cc, const quicly_loss_t *loss, int64_t now, uint32_t max_udp_payload_size);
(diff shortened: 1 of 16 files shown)


kazuho:
Thank you very much for the PR! I'll take a look and merge this.

kazuho:
Excuse me, based on my understanding that this PR only to change the two points listed below, I would close without doing anything:

Persistent congestion: it is not required by QUIC. RFC 9002 is merely "exemplary" ([Abstract of RFC 9000](https://datatracker.ietf.org/doc/html/rfc9000#abstract)), and [CUBIC](https://datatracker.ietf.org/doc/html/rfc9438) relies on QUIC PTO to handle the same case. Therefore, there is little need, if any, to implement it.

Use of unified loss detection rather than per-epoch states. This also is not a conformance issue, since loss detection is specified by RFC 9002. This is an intentional design choice and I do not see a compelling reason to adopt a more complex design.

I will post the same message to the original issue once I finish going through all of them.

sleepybishop:
Fair, thanks for taking a look.

kazuho:
Thank you so much for all your contributions. I highly appreciate them.

PR #691, opened 2026-09-05
Fix ubsan findings in pacer/frame

sleepybishop:
This PR fixes two undefined-behavior reports found via ubsan:

 - Avoid signed overflow when calculating the initial pacer window from the `INT64_MIN` reset sentinel.
 - Avoid null-pointer arithmetic when calculating `CONNECTION_CLOSE` frame capacity.


kazuho:
Thank you for the PR! Merging with tweaks.

PR #697, opened 2026-09-18
sub-millisecond RTT measurement and loss recovery

kazuho:
This PR retains sub-millisecond timing precision while keeping timer APIs in integer milliseconds, and opens the path to adopt congestion controllers that require fine-grained RTT observation.

* Timestamps and loss-recovery durations are represented using `double`; the unit of time remains the same: one millisecond.
* `quicly_context_t::now` provides current time using `double`; the change should be invisible to applications using the default `now` callback.
* `latest` and `minimum` of `quicly_rtt_t` are now `float`. This follows #680 that changed `smoothed` and `variance` to floats.
* `quicly_get_first_timeout` continues to return the timeout using `int64_t`, therefore applications can continue using their timer management code as is.
* The pacer continues to update its credit at millisecond granularity.

The last two are intentional. Millisecond-granularity timers and pacing avoid requiring frequent process wakeups. This is unlikely to cause performance issues, as ACKs are delivered to quicly as they arrive, allowing additional packets to be sent immediately when conditions permit.


include/quicly/sentmap.h
```diff
@@ -41,9 +41,9 @@ typedef struct st_quicly_sent_packet_t {
      */
     uint64_t packet_number;
     /**
-     *
+     * time at which the packet was sent (or INFINITY for the end-of-iteration sentinel)
      */
-    int64_t sent_at;
+    double sent_at;
```
bkatcher, line 46: Would it be more efficient to keep this (and now) as int64_t but just representing us instead of ms?

kazuho: On modern superscalar, out-of-order CPUs, floating-point arithmetic can potentially be faster. The protocol stack is generally integer-heavy, so throughput could improve by moving some calculations to otherwise underutilized floating-point execution units. quicly would only be used on CPUs powerful enough to handle encryption, which would likely also have floating-point execution units. Even without those units, the cost of floating-point emulation might still be negligible relative to the cost of encryption.

Therefore, my preference is to use one unified unit of time (i.e., milliseconds), using floating-point representations where we need finer-grained values. Floats are already used in quicly_rtt_t; this PR just uses them more, though it is of course fair to question whether the earlier changes to quicly_rtt_t were the right call.

Does that address your concern?

bkatcher: Yes, that's fine. Thank you.

